PR #3, opened 2021-09-20
PLAT-11326: return errors as JSON and others mini fixes

symphony-youri:
So we can eventually add error codes later, also easier to parse and
consistent with Spring errors.

---

Better deployment log 

Otherwise only the id (uuid) is printed.

---

Fix JSON schema for send-message.to.user-ids

Validation was broken with the empty enum.
Also do not assume how the args fields are named and use a placeholder.

### DIFF
--- a/workflow-language/src/main/resources/swadl-schema-1.0.json
+++ b/workflow-language/src/main/resources/swadl-schema-1.0.json
@@ -1406,7 +1406,7 @@
                             "enum": [
                                 "${variables.VARIABLE_NAME}",
                                 "${text(ACTIVITY_ID.outputs.message.message)}",
-                                "${event.args.content}",
+                                "${event.args.NAME}",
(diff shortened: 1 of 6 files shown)


workflow-language/src/main/resources/swadl-schema-1.0.json
```diff
@@ -2622,19 +2622,12 @@
         },
         "user-ids-inner": {
             "description": "User identifiers list.",
-            "anyOf": [
```
symphony-soufiane, line 2625: Do you still have auto completion with this?

symphony-soufiane (approved):

workflow-language/src/main/resources/swadl-schema-1.0.json
symphony-youri: Yes

PR #8, opened 2021-09-21
Fixed bug to check form-replied in one-of

symphony-soufiane:
The method checking if an activity is a formReplied was only checking events in the on list. When a form-replied event is put as part of an one-of event, the activity was not considered as a formReplied one

The PR is not complete as unit tests are not passing yet, this requires more investigation on different events. I will go back on this later.


symphony-youri:
Given how form replies are implemented (with subprocesses), is it possible to mix them with other events?

PR #9, opened 2021-09-21
Support blast messages in send-message

symphony-youri:
We can now send a message to multiple rooms with a single call to the
send-message activity (i.e. blast).


symphony-youri:
We might wonder about remove the stream-id and keep only stream-ids

symphony-soufiane (approved):

PR #12, opened 2021-09-22
PLAT-11280: Add get-attachment activity

symphony-soufiane:
A new activity has been added to get one attachment given message and attachment ids. Both attributes are required for this activity.
The activity downloads the attachment, decodes it and stores it in file system at ./workflows folder. Thus, the activity does not have the same output as the api. While this latter returns the attachment content as response, the activity stores it locally and returns the path.
In order to get the ./workflows folder path, a new functional interface ResourceFolderProvider was added since ResourceProvider could not have a second abstract method. The good practice is to keep only one abstract method in a functional interface.
The integration test with the working scenario results in file created in resources, at the end of the test we check that the file is present and we delete it. When it is not deleted, only a log is printed and this does not impact the test result passed/failed.
The reference documentation has been updated as well.

Finally, there is some duplicated code between GetAttachment and SendMessage part that handles attachments, this can be refactored in a next commit. The same for tests as well.


workflow-bot-app/src/test/java/com/symphony/bdk/workflow/GetAttachmentIntegrationTest.java
```diff
@@ -0,0 +1,141 @@
...
+    engine.deploy(workflow);
+
+    engine.onEvent(messageReceived("/get-attachment"));
+
+    verify(messageService, timeout(5000).times(1)).getMessage(msgId);
+    verify(messageService, timeout(5000).times(1)).getAttachment(streamId, msgId, attachmentId);
+
+    assertThat(workflow).isExecuted()
+        .hasOutput(String.format(OUTPUTS_ATTACHMENT_PATH_KEY, "getattachment"), expectedFilePath);
+    assertFileExistsAndDeleteIt(expectedFilePath);
```
symphony-youri, line 62: we should avoid using the file system while running tests
what we could do instead if have a store(InputStream) method in the resourceprovider interface and in our tests store it in memory so it can easily be retrieved and checked

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/ResourceFolderProvider.java
```diff
@@ -0,0 +1,5 @@
+package com.symphony.bdk.workflow.engine;
+
+public interface ResourceFolderProvider {
```
symphony-youri, line 3: we haven't really defined the other ResourceProvider as a functional interface, so I would probably at a method to it instead

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/attachment/GetAttachmentExecutor.java
```diff
@@ -0,0 +1,83 @@
...
+
+
+    String attachmentName = execution.getResourcesFolder() + attachmentInfo.getName();
+    downloadAndStoreAttachment(attachmentName, actualMessage.getStream().getStreamId(), actualMessage.getMessageId(),
+        attachmentInfo.getId(), execution.bdk().messages());
+
+    execution.setOutputVariable(OUTPUT_ATTACHMENT_PATH_KEY, attachmentName);
+  }
+
+  private void downloadAndStoreAttachment(String attachmentName, String streamId, String messageId,
```
symphony-youri, line 51: what if multiple workflows are downloading attachments with the same name?

symphony-youri: overwriting it maybe does not seem like a good idea

docs/reference.md
```diff
@@ -845,6 +845,22 @@ message returned.
 
 Example: _2021-08-31T15:50:00Z_
 
+### get-attachment
+
+Downloads an attachment to file system and returns its path.
```
symphony-youri, line 850: we should highlight that we do not clean the file and that it is up to the workflow. writer to do it
Or should we delete them when the JVM exists?

workflow-language/src/main/resources/swadl-schema-1.0.json
```diff
@@ -111,6 +111,18 @@
                             }
                         ]
                     },
+                    "get-attachment": {
+                        "description": "Get one attachment and stores it locally.",
```
symphony-youri, line 115: store it? (or gets?)

workflow-language/src/main/resources/swadl-schema-1.0.json
```diff
@@ -1459,6 +1471,18 @@
...
+            "type": "object",
+            "properties": {
+                "message-id": {
+                    "$ref": "#/definitions/message-id-inner"
+                },
+                "attachment-id": {
+                    "$ref": "#/definitions/single-attachment"
+                }
+            },
+            "required": ["message-id","attachment-id"]
```
symphony-youri, line 1484: file is not formatted

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/GetAttachmentIntegrationTest.java
symphony-soufiane: Actually we don't event need to store the files in tests. The method store(InputStream) stores the file and returns the its path which is the output of get-attachment , in the test we only returns the path without storing the file (assuming that the library method is working and does not need test)

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/configuration/WorkflowResourcesProvider.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @Override
+  public String saveResource(String name, byte[] content) {
+    String path = StringUtils.appendIfMissing(this.resourcesFolder, "/") + name;
+    File file = this.createFile(path);
+    return this.writeContent(file, content);
+  }
+
+  private File createFile(String name) {
```
symphony-youri, line 36: maybe there are functions in commons-io that do all of that for us

docs/reference.md
symphony-soufiane: done

symphony-youri (approved):

PR #13, opened 2021-09-23
PLAT-11401: Support variables usage for list/map attributes

symphony-youri:
In activity attributes with type list or map it was not possible to use
variable references (${}).

How are variables handled in workflows?

In SWADL you use them with ${}
When SWADL is deserialized to the Java model, strings were used
The Java model activity is serialized to JSON and stored in BPMN
When the BPMN is executed, expression (including variables) are
resolved and real values are replaced.
When the CamundaExecutor runs, the JSON activity is deserialised and the
Java mode rebuilt to be used in the Activity executor.

We now use a custom Jackson deserializer to be able to store in the Java
model either a resolved value or a variable reference (i.e the Variable
class).

We then use a custom TypeConverter during expression resolution (juel)
in Camunda to store complex objects (list,map) as escaped JSON.

When we deserialize to rebuild the Java model in CamundaExecutor another
custom deserializer is used to parse this escaped JSON and preserve the
complex objects.

Activities that have attributes other than strings should wrap them in
Variable<?>. Given JSON only support a generic number type, it is
difficult to differentiate a float/integer/long so it is up to the
activity developer to handle it.

The schema has been changed to allow to use variables or lists or maps
for the attributes that support it.

### DIFF
--- a/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/variable/EscapedJsonVariableDeserializer.java
+++ b/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/variable/EscapedJsonVariableDeserializer.java
@@ -1,0 +1,81 @@
+package com.symphony.bdk.workflow.engine.camunda.variable;
+
+import static com.symphony.bdk.workflow.swadl.v1.Variable.RESOLVED_VALUE_FIELD;
+import static com.symphony.bdk.workflow.swadl.v1.Variable.VARIABLE_REFERENCE_FIELD;
+
+import com.symphony.bdk.workflow.swadl.v1.Variable;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+
+/**
+ * Variables are serialized as escaped JSON by {@link VariableToJsonConverter} and passed to the executor.
+ * When the executor deserializes the activity to access the model, it uses this custom deserializer to
+ * recreate the {@link Variable} objects now containing resolved values.
+ */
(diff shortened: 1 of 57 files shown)


workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/variable/EscapedJsonVariableDeserializer.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Map;
+
+
+/**
+ * Variables are serialized as escaped JSON by {@link VariableToJsonConverter} and passed to the executor.
+ * When the executor deserializes the activity to access the model, it uses this custom deserializer to
+ * recreate the {@link Variable} objects now containing resolved values.
+ */
+@SuppressWarnings("rawtypes")
+public class EscapedJsonVariableDeserializer extends StdDeserializer<Variable> {
```
symphony-soufiane, line 26: Shouldn't this be moved to workflow/swadl/ like ActivityDeserializer?

symphony-youri: this used by CamundaExecutor and relate to how expressions are resolved (so specific to Camunda)

workflow-language/src/main/java/com/symphony/bdk/workflow/swadl/v1/Variable.java
```diff
@@ -0,0 +1,49 @@
...
+    this.resolvedValue = resolvedValue;
+  }
+
+  @JsonIgnore
+  public T get() {
+    return resolvedValue;
+  }
+
+  @JsonIgnore
+  public Integer getInt() {
```
symphony-soufiane, line 41: why only getInt() ? are the other types all handled by the method above?

symphony-soufiane (approved):

workflow-language/src/main/java/com/symphony/bdk/workflow/swadl/v1/Variable.java
symphony-youri: It was the only one that i needed :) (but we could do the same for other types)
we use it for pagination/skip/limit
for user ids, it is often a list of long so the logic is a bit different and kept in the executors

symphony-elias (approved):

PR #14, opened 2021-09-23
Made use of all fields in create-room

symphony-soufiane:
In a previous commit, we added some fields that were missing in create-room activity such as memberCanInvite... Or, these fields were only added in the pojo and swadl-schema and were not used in the executor.
In this commit we made use of them in the executor and updated the tests


workflow-language/src/main/java/com/symphony/bdk/workflow/swadl/v1/activity/room/CreateRoom.java
```diff
@@ -48,5 +48,53 @@
   public Boolean isPublicAsBool() {
     return Boolean.valueOf(isPublic);
   }
+
```
symphony-youri, line 51: they can be replaced with Variable<Boolean> now

symphony-youri (approved):

PR #17, opened 2021-09-27
Handle multiple workflows/request-received-event

symphony-youri:
If multiple workflows are listening to request-received-event they would
all get triggered because the signal's name is the same for all
workflows.

Making the workflow id part of the signal name solves this issue.


symphony-soufiane (approved):
Thank you for the fix; LGTM.

PR #18, opened 2021-09-29
Support H2 persistent database

symphony-soufiane:
Previously, we were only supporting in-memory h2 database. Hence, workflows state and data was lost when the app restarts.
The datasource configuration was done in a bean as part of camunda engine configuration class. In this commit, we moved that configuration to application.yaml file to allow users to switch it to a persistent database.
The persistent database is an H2 file based one, it can be enabled by changing the database url in application.yaml file.

Deployment documentation was changed to explain that.

For the same occasion, Camunda bpmn builder was changed to not add suffix to workflows names/ids. Actually, using the persistent database, when we deploy a workflow that has already been deployed, if they have the same id, then only the second one is used for new executions. The pending executions use the old one when they are relaesed.
To support that, we had to change the builder to not add suffix. Thus, workflows can be versionned when a new one is deployed.


workflow-bot-app/src/main/resources/application.yaml
```diff
@@ -28,9 +28,15 @@ camunda:
 
 # Spring boot configuration
 spring:
+  datasource:
```
symphony-youri, line 31: are the " required here?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/bpmn/CamundaBpmnBuilder.java
```diff
@@ -110,7 +109,12 @@ private void checkParentIsKnown(Map<String, String> parentActivities, String wor
   }
 
   private BpmnModelInstance workflowToBpmn(Workflow workflow) throws JsonProcessingException {
-    ProcessBuilder process = Bpmn.createExecutableProcess(createUniqueProcessId(workflow)).name(workflow.getId());
+    String processId = "";
+    if (workflow.getId() != null) {
```
symphony-youri, line 113: if the workflow id is not set, is it ok to have a process with an empty id? do the logs look ok in that situation?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/bpmn/CamundaBpmnBuilder.java
```diff
@@ -110,7 +109,12 @@ private void checkParentIsKnown(Map<String, String> parentActivities, String wor
...
+      processId = workflow.getId().replaceAll("\\s+", "_");
```
symphony-youri, line 114: why do we do the replace all (we lost the comment here)

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/bpmn/CamundaBpmnBuilder.java
```diff
@@ -110,7 +109,12 @@ private void checkParentIsKnown(Map<String, String> parentActivities, String wor
...
+    }
+
+    ProcessBuilder process = Bpmn.createExecutableProcess(processId).name(workflow.getId());
```
symphony-youri, line 117: we can add a unit test for the workflow update, deploying 2 different workflows with the same id and checking that it is the latest deployed one that is executed

docs/deployment.md
```diff
@@ -35,14 +37,19 @@ properties are listed
...
 workflows, for instance the wait time to detect new events to process. It is configured with a low value by default to
 ensure the bot is reactive.
 
+#### Retry on activity/task errors
 Camunda is configured to retry on activity/task errors. Part of the error handling is done via the BDK that already
 support retrying on failed API calls and then error handling can also be done when writing workflows with
 the [activity-failed event](./reference.md#activity-failed).
 
+#### Persistent database
+In order to use a persistent database, you need to change `spring.datasource.url` to `"jdbc:h2:file:./data/process_engine;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=FALSE"`
```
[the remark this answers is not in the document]
symphony-youri: instead of process_engine we might want to put a "wdk" specific name (process engine won't speak to wdk users)

docs/deployment.md
```diff
@@ -5,8 +5,10 @@ the [BDK Spring Boot integration](https://symphony-bdk-java.finos.org/spring-boo
 The [Getting started](./getting-started.md) guide explains how the Symphony Generator can be used to create a sample
 project with configuration files.
 
-_In its current version, the WDK does not persist state except in memory, so restarting the bot means all running
-workflow's state is lost. Running multiple instances of the bot for high availability is also not yet supported._
+By default, the WDK does not persist state except in memory, so restarting the bot means all running
```
symphony-youri, line 8: do we want to make the usage of the h2 file DB the default? (not for unit tests) so that is easier for users to upgrade a bot?

docs/deployment.md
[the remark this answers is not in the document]
symphony-soufiane: The folder data/ is created when it does not exist. Corrected the doc

docs/deployment.md
symphony-soufiane: That's what was specified in the ticket and I think it makes sense as one would only need to test the workflow in the beginning. The database will be needed once we test the workflow and know how it works in order to make real things.

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/bpmn/CamundaBpmnBuilder.java
symphony-soufiane: I put back the comment.
Technically at this level we can't have a workflow without id. workflowToBpmn() method is called by addWorkflow() method that is called by CamundaEnigne.deploy(). The latter sets workflow id if not provided.

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/bpmn/CamundaBpmnBuilder.java
symphony-soufiane: I put back the comment.
Technically at this level we can't have a workflow without id. workflowToBpmn() method is called by addWorkflow() method that is called by CamundaEnigne.deploy(). The latter sets workflow id if not provided.

PR #24, opened 2021-10-04
PLAT-11371: Added curl activity

symphony-soufiane:
### Description
This PR is related to [PLAT-11371](https://perzoinc.atlassian.net/browse/PLAT-11371)
In this PR, we added a new activity send-request that can be used to perform HTTP requests. Its outputs are the request status and the response body if it succeed.
A dependency id added to include BDK http jeresey2.
We reuse the same bot configuration for api client, so no additional configuration is needed. This way, the proxy and other settings already set can be used.
The api client is supposed to be used to target different apis and not necessarily the pod. Hence, the host url changes in every request. Thats why the apiClient is reinitialized with a new basepath for each execution. The response body is stored in outputs and we dont expect to have a big amount of data returned by the api. In this case, the disk could be full.

The reference documetation and integration tests have been updated for this occasion.

### Checklist
- [x] Javadoc added or updated
- [x] Updated the documentation in [docs folder](../docs)

### DIFF
--- a/workflow-bot-app/build.gradle
+++ b/workflow-bot-app/build.gradle
@@ -9,7 +9,7 @@ dependencies {
     implementation platform('org.finos.symphony.bdk:symphony-bdk-bom:2.3.0')
 
     implementation 'org.finos.symphony.bdk:symphony-bdk-core-spring-boot-starter'
-    implementation 'org.finos.symphony.bdk:symphony-bdk-http-jersey2'
+    implementation 'org.springframework.boot:spring-boot-starter-webflux'
     implementation 'org.springframework.boot:spring-boot-starter-web'
     implementation 'org.springframework.boot:spring-boot-starter-data-jdbc'
     implementation 'org.springframework.boot:spring-boot-starter-actuator'
--- a/workflow-bot-app/build.gradle
+++ b/workflow-bot-app/build.gradle
@@ -9,7 +9,6 @@ dependencies {
     implementation platform('org.finos.symphony.bdk:symphony-bdk-bom:2.3.0')
 
     implementation 'org.finos.symphony.bdk:symphony-bdk-core-spring-boot-starter'
-    implementation 'org.springframework.boot:spring-boot-starter-webflux'
     implementation 'org.springframework.boot:spring-boot-starter-web'
     implementation 'org.springframework.boot:spring-boot-starter-data-jdbc'
     implementation 'org.springframework.boot:spring-boot-starter-actuator'
--- a/workflow-bot-app/build.gradle
+++ b/workflow-bot-app/build.gradle
@@ -9,6 +9,7 @@ dependencies {
     implementation platform('org.finos.symphony.bdk:symphony-bdk-bom:2.3.0')
 
     implementation 'org.finos.symphony.bdk:symphony-bdk-core-spring-boot-starter'
+    implementation 'org.finos.symphony.bdk:symphony-bdk-http-jersey2'
     implementation 'org.springframework.boot:spring-boot-starter-web'
     implementation 'org.springframework.boot:spring-boot-starter-data-jdbc'
     implementation 'org.springframework.boot:spring-boot-starter-actuator'
--- a/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/client/ApiClientFactory.java
+++ b/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/client/ApiClientFactory.java
@@ -1,33 +1,19 @@
 package com.symphony.bdk.workflow.client;
 
-import com.symphony.bdk.core.config.model.BdkConfig;
+import com.symphony.bdk.core.util.ServiceLookup;
 import com.symphony.bdk.http.api.ApiClient;
 import com.symphony.bdk.http.api.ApiClientBuilder;
-import com.symphony.bdk.http.jersey2.ApiClientBuilderProviderJersey2;
-
-import org.apache.commons.lang3.StringUtils;
+import com.symphony.bdk.http.api.ApiClientBuilderProvider;
 
 public class ApiClientFactory {
 
-  private final BdkConfig bdkConfig;
+  private final ApiClientBuilder apiClientBuilder;
 
-  private ApiClientBuilder apiClientBuilder;
-
-  public ApiClientFactory(BdkConfig bdkConfig) {
-    this.bdkConfig = bdkConfig;
-    final ApiClientBuilderProviderJersey2 apiClientBuilderProvider = new ApiClientBuilderProviderJersey2();
+  public ApiClientFactory() {
+    final ApiClientBuilderProvider apiClientBuilderProvider =
+        ServiceLookup.lookupSingleService(ApiClientBuilderProvider.class);
 
     this.apiClientBuilder = apiClientBuilderProvider.newInstance();
-
-    if (this.bdkConfig.getProxy() != null) {
-      this.apiClientBuilder =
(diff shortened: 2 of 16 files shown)


workflow-bot-app/src/test/java/com/symphony/bdk/workflow/IntegrationTest.java
```diff
@@ -118,6 +123,10 @@ protected static Stream stream(String streamId) {
   @BeforeEach
```
symphony-soufiane, line 123: After first commit, tests were broken because the injected bdk services were null. 
I came up with this solution to fix the tests, but I'm not convinced it is the best way. see my commit message here https://github.com/finos/symphony-wdk/pull/24/commits/715252cfbe2a3124a812432ab96cf8db5bcfeabe

workflow-bot-app/build.gradle
```diff
@@ -9,6 +9,7 @@ dependencies {
     implementation platform('org.finos.symphony.bdk:symphony-bdk-bom:2.3.0')
 
     implementation 'org.finos.symphony.bdk:symphony-bdk-core-spring-boot-starter'
+    implementation 'org.finos.symphony.bdk:symphony-bdk-http-jersey2'
```
symphony-thibault, line 12: As the workflow bot is relying on the Spring Boot Starter, shouldn't we use the `webclient` implementation instead of `jersey2`?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/client/ApiClientFactory.java
```diff
@@ -0,0 +1,37 @@
+package com.symphony.bdk.workflow.client;
+
+import com.symphony.bdk.core.config.model.BdkConfig;
+import com.symphony.bdk.http.api.ApiClient;
+import com.symphony.bdk.http.api.ApiClientBuilder;
+import com.symphony.bdk.http.jersey2.ApiClientBuilderProviderJersey2;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class ApiClientFactory {
```
symphony-thibault, line 10: Don't we need to support truststore configuration as well? (cf. https://github.com/finos/symphony-bdk-java/blob/main/symphony-bdk-core/src/main/java/com/symphony/bdk/core/client/ApiClientFactory.java)

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/RequestExecutor.java
```diff
@@ -0,0 +1,43 @@
...
+      ApiResponse<Object> apiResponse = execution.bdk()
+          .apiClient(activity.getUrl())
+          .invokeAPI("", activity.getMethod(), Collections.emptyList(), activity.getBody(), activity.getHeaders(),
+              Collections.emptyMap(), Collections.emptyMap(), "application/json", "application/json", null,
+              new TypeReference<>() {});
+
+      data = apiResponse.getData();
+      statusCode = apiResponse.getStatusCode();
+    } catch (ApiException apiException) {
+      statusCode = apiException.getCode();
```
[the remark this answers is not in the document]
symphony-soufiane: The error message from apiException for example? 

I considered that the status code was enough for the subsequent activities. The error stack will be shown in the console any way for the developer.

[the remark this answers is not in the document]
symphony-thibault: Not only, the whole `apiException.getResponseBody()` that is a JSON object as well

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/client/ApiClientFactory.java
```diff
@@ -0,0 +1,37 @@
...
+
+  private final BdkConfig bdkConfig;
+
+  private ApiClientBuilder apiClientBuilder;
+
+  public ApiClientFactory(BdkConfig bdkConfig) {
+    this.bdkConfig = bdkConfig;
+    final ApiClientBuilderProviderJersey2 apiClientBuilderProvider = new ApiClientBuilderProviderJersey2();
```
symphony-thibault, line 18:
```suggestion
    final ApiClientBuilderProvider apiClientBuilderProvider = ServiceLookup. lookupSingleService(ApiClientBuilderProvider.class);
```

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/ExecuteRequestIntegrationTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+  @Test
+  void executeRequestSuccessful() throws Exception {
+    final Workflow workflow =
+        SwadlParser.fromYaml(getClass().getResourceAsStream("/request/execute-request.swadl.yaml"));
+
+    final ApiClient mockedApiClient = mock(ApiClient.class);
+    final Map<String, String> header = Map.of("headerKey", "headerValue");
+    final Map<String, Object> body = Map.of("args", Map.of("key", "value"));
+    final ApiResponse<Object> mockedResponse = new ApiResponse<>(200, Collections.emptyMap(), "OK");
```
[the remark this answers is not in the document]
symphony-soufiane: To process the result we can go through the JSON nodes.
For example, this api returns exchange rates in JSON ([https://open.er-api.com/v6/latest/EUR](https://open.er-api.com/v6/latest/EUR)). 
If we want the rate for "EUR" we can have it with `${ACTIVITY_ID.outputs.body.rates['EUR']}`

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/ExecuteRequestIntegrationTest.java
```diff
@@ -0,0 +1,76 @@
...
+        eq("application/json"), eq("application/json"), eq(null), any(TypeReference.class)))
+        .thenReturn(mockedResponse);
+
+    engine.deploy(workflow);
+
+    engine.onEvent(messageReceived("/execute"));
+
+    assertThat(workflow).isExecuted()
+        .hasOutput(String.format(OUTPUTS_STATUS_KEY, "executeGetRequest"), 200)
+        .hasOutput(String.format(OUTPUTS_BODY_KEY, "executeGetRequest"), jsonResponse);
```
symphony-thibault, line 53: Could you also unit test that you can retrieve a specific field of the response? Following the swadl example: 
```yaml
id: exchange-rates-workflow
activities:
  - execute-request:
      id: callExchangeRatesApi
      method: GET
      url: https://open.er-api.com/v6/latest/EUR
      on:
        message-received:
          content: /show-exchange-rates

  - send-message:
      id: sendRates
      content: EURO= ${callExchangeRatesApi.outputs.body.rates['EUR']}, USD= ${callExchangeRatesApi.outputs.body.rates['USD']}
```

here we would ensure that it's possible to retrieve the `name` value

symphony-soufiane: There was nothing to do when the request succeeds. When an ApiException is catched, I map the responseBody (which is a string) to a map before storing it in outputs. Thus, we can process it in SWADL.
For tests, I added a script activity that checks the output values (...body.contact.phone and others). It makes the test fail if the actual value is not expected or not found.

symphony-thibault (approved):
LGTM 👍

PR #58, opened 2021-10-27
PLAT-11536: Support any content type in execute-request activity

symphony-soufiane:
An API can return a simple String as exception message (not key/value format). Since this is considered as a valid JSON, we have to support it. In this case, we create a key (message) and put the string inside before setting it in activity outputs.

Documentation updated to mention that timeout is set to 24H by default for form-replied event, and no default for other activities.


workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/RequestExecutor.java
```diff
@@ -42,29 +37,52 @@ public void execute(ActivityExecutorContext<ExecuteRequest> execution) throws Js
...
+          this.httpClient.execute(activity.getMethod(), activity.getUrl(), activity.getBody(),
+              headersToString(activity.getHeaders()));
+
+      statusCode = response.getCode();
+
+
+      if (HttpStatus.valueOf(statusCode).is2xxSuccessful()) {
+        data = response.getContent();
+      } else {
+        // we need to set the responseBody as a map instead of a string to allow processing it in subsequent activities
```
symphony-youri, line 50: why?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/RequestExecutor.java
```diff
@@ -42,29 +37,52 @@ public void execute(ActivityExecutorContext<ExecuteRequest> execution) throws Js
...
+        data = this.handleException(response.getContent());
+        log.debug("This error happens when the request fails. {}", response.getContent());
       }
-      log.debug("This error happens when the request fails.", apiException);
+
+    } catch (IOException ioException) {
+      statusCode = 500;
+      data = this.handleException(ioException.getMessage());
```
symphony-youri, line 57: in that case I would say we probably want to fail the execution of the activity

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/client/HttpClient.java
```diff
@@ -0,0 +1,96 @@
...
+  private HttpResponse executeWithoutBody(Request request) throws IOException {
+    return request.execute().returnResponse();
+  }
+
+  private HttpResponse executeWithBody(Request request, Object body, Map<String, String> headers)
+      throws IOException {
+
+    // set body
+    String contentType = headers.get(CONTENT_TYPE);
+    if (ContentType.MULTIPART_FORM_DATA.getMimeType().equals(contentType)) {
```
symphony-youri, line 81: can't we have the body as an object in the model and then if the content type is form, enforce it to be a map directly? instead of passing a JSON map inside YAML

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/UtilityFunctionsMapper.java
```diff
@@ -20,18 +22,23 @@
...
+    FUNCTION_MAP.put("json", ReflectUtil.getMethod(UtilityFunctionsMapper.class, "json", String.class));
+    FUNCTION_MAP.put("escape", ReflectUtil.getMethod(UtilityFunctionsMapper.class, "escape", String.class));
+    FUNCTION_MAP.put("mentions", ReflectUtil.getMethod(UtilityFunctionsMapper.class, "mentions", Object.class));
+    FUNCTION_MAP.put("hashTags", ReflectUtil.getMethod(UtilityFunctionsMapper.class, "hashTags", Object.class));
+    FUNCTION_MAP.put("cashTags", ReflectUtil.getMethod(UtilityFunctionsMapper.class, "cashTags", Object.class));
+    FUNCTION_MAP.put("emojis", ReflectUtil.getMethod(UtilityFunctionsMapper.class, "emojis", Object.class));
+  }
+
+  public static Map json(String string) throws JsonProcessingException {
+    return new ObjectMapper().readValue(string, Map.class);
```
symphony-youri, line 41: creating ObjectMapper is a bit costly, we usually try to avoid that (i.e putting it as a static variable)
what if the JSON is a string or an array (which is valid JSON)? would not we be better off returning a JsonNode? (but leaking jackson API inside workflow)

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/client/HttpClient.java
symphony-soufiane: To be able to enforce cast to a map, the body should not be a String. So in case of multipart-form-data, if we provide the body as an object it's fine. 
However, this doesn't work for other content types where we need to provide the body in a given format (in JSON for application/json, and a different format for text/plain). So to write it in the required format, we need a String. 
In next commit, I'm changing json-schema to define the body as string and object, and it is object in the java model. Like this we can support both: when we use multipart type, we provide an object, and a string for other types. This can be ambiguous though.

docs/reference.md
```diff
@@ -1826,35 +1828,59 @@ Allowed values:
...
+```yaml
+activities:
+  - execute-request:
+      id: myRequest
       headers:
-        "X-Workflow-Token": "A_TOKEN"
+        Content-Type: multipart/form-data
+        Accept: multipart/form-data
+      body:
+        outer:
```
symphony-youri, line 1869: does it really work with a nested structure? or does it support only one level (i.e just key-value)

docs/reference.md
```diff
@@ -1826,35 +1828,59 @@ Allowed values:
...
 [method](#method) | String | Yes |
-[body](#body) | String | No |
+[body](#body) | Object/String | No |
 [headers](#headers) | String | No |
 
+_nb: For multipart/form-data content type requests, the body should be provided as a key/value object. For other content types, it can be provided as String in JSON format.
+
+Output | Type |
+----|----|
+body | String containing the response body
```
symphony-youri, line 1842: To be updated?

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/utils/UtilityFunctionsMapperTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+
+@SuppressWarnings("unchecked")
+class UtilityFunctionsMapperTest {
+
+  @Test
+  void testString() {
```
symphony-youri, line 15: maybe add json as part of the method's name? (because we have other methods in this class too)

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/utils/UtilityFunctionsMapperTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+  @Test
+  void testInteger() {
+    final Object json = UtilityFunctionsMapper.json("2");
+    assertThat(json).isEqualTo(2);
+  }
+
+  @Test
+  void testBoolean() {
+    final Object json = UtilityFunctionsMapper.json("false");
```
symphony-youri, line 28: In that case will we receive true/false/2 (not strings) or strings?

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/ExecuteRequestIntegrationTest.java
```diff
@@ -1,84 +1,135 @@
...
 
     engine.deploy(workflow);
 
     engine.onEvent(messageReceived("/execute-failed"));
 
     assertThat(workflow).isExecuted()
         .hasOutput(String.format(OUTPUTS_STATUS_KEY, "executeGetRequest"), 400)
-        .hasOutput(String.format(OUTPUTS_BODY_KEY, "executeGetRequest"),
-            Map.of("message", "ApiException response body"));
+        .hasOutput(String.format(OUTPUTS_BODY_KEY, "executeGetRequest"), exceptionMessage);
```
symphony-youri, line 112: aren't we supposed to receive a map in one case?

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/ExecuteRequestIntegrationTest.java
```diff
@@ -1,84 +1,135 @@
...
+    when(httpClient.execute("GET", url, null, header)).thenReturn(mockedResponse);
 
     engine.deploy(workflow);
 
     engine.onEvent(messageReceived("/execute"));
 
     assertThat(workflow).isExecuted()
         .hasOutput(String.format(OUTPUTS_STATUS_KEY, "executeGetRequest"), 200)
-        .hasOutput(String.format(OUTPUTS_BODY_KEY, "executeGetRequest"), jsonResponse);
+        .hasOutput(String.format(OUTPUTS_BODY_KEY, "executeGetRequest"), expectedResponse);
```
symphony-youri, line 84: isn't it supposed to be a map?

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/SendMessageIntegrationTest.java
```diff
@@ -33,6 +33,7 @@
 import java.util.Collections;
 import java.util.List;
 
+@SuppressWarnings("unchecked")
```
symphony-youri, line 36: We can also put the annotation directly on the var declaration to avoid it being applied everywhere

workflow-language/src/main/resources/swadl-schema-1.0.json
```diff
@@ -1546,17 +1547,24 @@
...
-                    "type": "object",
-                    "description": "HTTP request body"
+                    "oneOf": [
+                        {
+                            "type": "string",
+                            "description": "HTTP request body"
+                        },
+                        {
+                            "type": "object",
+                            "description": "HTTP request body"
```
symphony-youri, line 1557: we can explain here it works for a specific content type

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/RequestExecutor.java
```diff
@@ -1,49 +1,62 @@
...
 
-    execution.setOutputVariable(OUTPUT_STATUS_KEY, statusCode);
-    execution.setOutputVariable(OUTPUT_BODY_KEY, data);
+  private Map<String, String> headersToString(Map<String, Object> headers) {
+    Map<String, String> result = new HashMap<>();
+    for (Map.Entry<String, Object> entry : headers.entrySet()) {
+
+      // We consider that opening/closing brackets are not allowed in headers values.
+      // Otherwise, the list items should be added in the string one by one using map function
+      String entryListAsString = List.of(entry.getValue()).toString();
```
symphony-youri, line 54: is it covered by tests?

can't we support a map of arrays if we want to support repeated headers? (to avoid parsing the strings)

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/RequestExecutor.java
```diff
@@ -1,49 +1,62 @@
...
-      data = new ObjectMapper().readValue(apiException.getResponseBody(), Map.class);
-      log.debug("This error happens when the request fails.", apiException);
+      Response response =
+          this.httpClient.execute(activity.getMethod(), activity.getUrl(), activity.getBody(),
+              headersToString(activity.getHeaders()));
+
+      execution.setOutputVariable(OUTPUT_STATUS_KEY, response.getCode());
+      execution.setOutputVariable(OUTPUT_BODY_KEY, response.getContent());
+    } catch (IOException ioException) {
+      log.debug("This error happens when the request fails with IOException.", ioException);
```
symphony-youri, line 43: do we need to log here?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/RequestExecutor.java
```diff
@@ -1,49 +1,62 @@
...
+      throw new RuntimeException(ioException.getMessage());
```
symphony-youri, line 44: Exceptions should be wrapped so we have a proper stacktraces, here we are loosing the original exception (new RunTimeException(ioException))

symphony-youri: beside we already throw ioexception from the executors, do we need to catch it then? (for all BDK calls right now we let the ioexception be raised)

workflow-language/src/main/java/com/symphony/bdk/workflow/swadl/v1/activity/request/ExecuteRequest.java
```diff
@@ -16,7 +16,7 @@
 @Data
 public class ExecuteRequest extends BaseActivity {
   private String url;
-  private String method;
-  private Map<String, Object> body = Collections.emptyMap();
-  private Map<String, String> headers = Collections.emptyMap();
+  private String method = "GET";
```
symphony-youri, line 19: should it be an enum?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/client/HttpClient.java
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.hc.core5.http.HttpResponse;
+import org.springframework.stereotype.Component;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+@Generated
+@SuppressWarnings("unchecked")
```
symphony-youri, line 18: this can be set at method level too

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/client/HttpClient.java
```diff
@@ -0,0 +1,96 @@
...
+@Component
+public class HttpClient {
+
+  private static final String CONTENT_TYPE = "Content-Type";
```
symphony-youri, line 22: HttpHeaders.CONTENT_TYPE can be used

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/client/HttpClient.java
```diff
@@ -0,0 +1,96 @@
...
+
+  public Response execute(String method, String url, Object body, Map<String, String> headers)
+      throws IOException {
+    HttpResponse httpResponse;
+    Request request;
+    switch (method) {
+      case "GET":
```
symphony-youri, line 29: there is a Request.create method that can replace the switch

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/client/HttpClient.java
```diff
@@ -0,0 +1,96 @@
...
+        break;
+      case "OPTIONS":
+        request = Request.options(url);
+        httpResponse = this.executeWithBody(request, body, headers);
+        break;
+      default:
+        throw new RuntimeException("Method is not supported");
+    }
+
+    ClassicHttpResponse classicHttpResponse = (ClassicHttpResponse) httpResponse;
```
symphony-youri, line 61: are we sure we can always cast it? should not we be more defensive?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/client/HttpClient.java
```diff
@@ -0,0 +1,96 @@
...
+      throws IOException {
+
+    // set body
+    String contentType = headers.get(CONTENT_TYPE);
+    if (ContentType.MULTIPART_FORM_DATA.getMimeType().equals(contentType)) {
+      Form form = Form.form();
+      Map<String, Object> bodyAsMap = (LinkedHashMap<String, Object>) body;
+      bodyAsMap.forEach((key, value) -> form.add(key, value.toString()));
+      request.bodyForm(form.build());
+    } else if (!contentType.isEmpty()) {
```
symphony-youri, line 86: perhaps we should document that the content-type is excepted if a body is set?

docs/reference.md
symphony-soufiane: What should I set?  This didn't change, outputs.body still has the response body, in a String format.

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/utils/UtilityFunctionsMapperTest.java
symphony-soufiane: **"json instanceof String"** returns false, **"json instanceof Boolean"** returns true.
Does it worth to add it to the tests?

workflow-language/src/main/java/com/symphony/bdk/workflow/swadl/v1/activity/request/ExecuteRequest.java
symphony-soufiane: It could be an enum but I don't think we need that in java model since the json-schema verifies the value at SWADL level (it is defined as an enum in the schema).

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/client/HttpClient.java
symphony-soufiane: Do we throw an exception here if the body is set without a content-type? 
We can also check that dependency in json-schema.
Or we just ignore the body in this case?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/RequestExecutor.java
symphony-soufiane: In ExecuteRequestIntegrationTest.java: 

**SWADL with a header having a list of items:** https://github.com/finos/symphony-wdk/pull/58/files#diff-57f7d5d71b9c4de5368b1d6e2e033fda26423133a9b0ebd5f20694ed9e21c0eeR15
**Expected map to be used:**  https://github.com/finos/symphony-wdk/pull/58/files#diff-26eea4bd568d90ac0dc9e316804f84a2f6f7d01bea0f8fb2d1b017e5b0dc7ce4R43

symphony-soufiane: If we consider headers as a map of arrays then a header that only has 1 value is considered as a list of 1 item, or am I wrong? 
In this case we will have to set headers in SWADL as a list (of 1 item when not repeated).

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/ExecuteRequestIntegrationTest.java
symphony-soufiane: If the api returns "This is a regular String" as response, what do we put as key if it is a map?
Didn't we agree to put exactly what the api returns as a String, that can be a JSON (eventually a String/Integer which are valid JSON) or anything else. The utility function is used to process that as JSON in subsequent activities.

symphony-youri: my bad I'm mixing when json() is used or nut

docs/reference.md
symphony-youri: That's me, mixing json() usage and the raw body

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/utils/UtilityFunctionsMapperTest.java
symphony-youri: I was expecting the test to be 
```     final Object json = UtilityFunctionsMapper.json(false);```
But I' mixing json() and the raw response again

File: workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/client/HttpClient.java (hunk 21 above)
Comment (symphony-youri) [reply workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/client/HttpClient.java]: I would perhaps use a default content type in this case (json?) and set the body
In Apache HTTP client for org.apache.hc.core5.http.io.entity.HttpEntities#create(java.lang.String) the default is text plain but I'm not sure it makes sense when calling APIs

File: workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/RequestExecutor.java (hunk 13 above)
Comment (symphony-youri) [reply workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/RequestExecutor.java:47]: Good, we have tests for it and the YAML looks good, but then why do we have to parse/remove []?
Can't we just take all the values and join them if they are array items?
it looks like what we do relies on the .toString() of a list (that might not be a list), it seems fragile

File: workflow-language/src/main/java/com/symphony/bdk/workflow/swadl/v1/activity/request/ExecuteRequest.java (hunk 16 above)
Comment (symphony-youri) [reply workflow-language/src/main/java/com/symphony/bdk/workflow/swadl/v1/activity/request/ExecuteRequest.java:19]: but using an enum makes it cleaner in the code (especially when we had the switch)
now without the switch keeping a string is probably simpler (but could be error prone)

File: workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/RequestExecutor.java (hunk 13 above)
Comment (symphony-soufiane) [reply workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/RequestExecutor.java:47]: `it looks like what we do relies on the .toString() of a list (that might not be a list), it seems fragile`:
I enforce it to be a list. _.toString()_ is called on _List.of(header).toString()_ ==> So in case the header is not a list we enforce it to be so before calling toString()

`Can't we just take all the values and join them if they are array items?`
By using "instance of"? so we join them only if the object is an instance of list? I tried to avoid using that.

The issue here is that String.joins() requires a CharSequence parameter and we cannot cast List -> CharSequence. 
I agree that this seems fragile,

Comment (symphony-soufiane) [reply workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/RequestExecutor.java:47]:
```
if (entry.getValue() instanceof List) {
      StringBuilder stringBuilder = new StringBuilder();
      ((List) entry.getValue()).forEach(item -> stringBuilder.append(item).append(","));

      result.put(entry.getKey(), String.join(",", stringBuilder.deleteCharAt(stringBuilder.length() - 1)));
} else {
      result.put(entry.getKey(), (String) entry.getValue());
}
```

Comment (symphony-soufiane) [reply workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/RequestExecutor.java:47]: WDYT?  (I was avoiding to use instance of but I don't see a better way)

```
  private Map<String, String> headersToString(Map<String, Object> headers) {
    Map<String, String> result = new HashMap<>();
    for (Map.Entry<String, Object> entry : headers.entrySet()) {
      StringBuilder stringBuilder = new StringBuilder();
      toList(entry.getValue()).forEach(item -> stringBuilder.append(item).append(","));
      result.put(entry.getKey(), stringBuilder.deleteCharAt(stringBuilder.length() - 1).toString());
    }
    return result;
  }

  @SuppressWarnings("unchecked")
  private List<String> toList(Object object) {
    if (object instanceof String) {
      return List.of((String) object);
    } else {
      return new ArrayList<>(((List<String>) object));
    }
  }
```

symphony-youri:
I would maybe suggest to rebase some of the commits so we a better view of the end results

symphony-youri (approved):
I do not if we have users already using the execute-request activity but let's maybe communicate (at least internally) about the changes

PR #62, opened 2021-11-09
Fix variable deserialization for lists 

### LINKED ISSUE #61: Allow variable usage of variables in list users-id
user-ids parameter requires a list of Number-based ids if we try to use ${...} the ids are passed as String and the workflow fails. We should allow the usage of variables in this field.

Example of workflow: 

```
- send-message:
  id: informationAnalysis
  on:
    form-replied:
      form-id: requestApprover
  to:
    user-ids:
      - ${event.initiator.user.userId}
  content: |
    Please review the following Information Analysis
```

symphony-youri:
Variables that are lists were supported, however list of variables were
not. We now support both by declaring the activity attributes as
Variables of List of Variables.

Ser/Deser has been updated to reflect that. We do not consider this a
perfect solution for now as it makes the activity writing a bit
cumbersome. Also maps are not fully supported here.

Fixes #61 PLAT-11547


symphony-mariacristina (approved):
LGTM!

symphony-soufiane (approved):

PR #63, opened 2021-11-10
Simplify variable handling in activities

symphony-youri:
This is another take at #61.

Here we dropped the Variable class and kept only the classic types in
activities. To support variable replacement we serialize/deserialize the
activity in different ways at each step of the workflow
construction/deployment/execution.

Check the comments in BaseActivity for more details.


symphony-mariacristina (approved):
LGTM!

symphony-soufiane (approved):
I reviewed, mostly to be aware of the changes :)

PR #64, opened 2021-11-16
PLAT-11243 Support external message template

symphony-mariacristina:
### Description
Goal of this implementation is to support external message template while sending/updating a message.
We will now be able to specify the content of a message by:
- adding directly the content in the swadl file
```
	send-message:
	  content: This is the content of the message
```
- adding the message content as an external xml file
```
	send-message:
	  content:
	    template: message.mml.xml
```
- using a freemarker file, in this case the list of variables available in the workflow will be passed along with the template so that we can use variables inside the ftl file and get the match between the existing variables
```
	send-message:
	  content:
	    template: messageWithFreemarker.ftl
```

### What is missing?
- [x] for now the path has to be "./workflows/message.mml.xml" -> to be fixed with only "message.mml.xml"
- [x] Unit testing
- [x] Swadl schema update
- [x] Documentation update
- [ ] Release new BDK version


workflow-language/src/main/java/com/symphony/bdk/workflow/swadl/v1/activity/message/SendMessage.java
```diff
@@ -13,7 +12,7 @@
 @EqualsAndHashCode(callSuper = true)
 @Data
 public class SendMessage extends BaseActivity {
-  private String content;
+  private Object content;
```
symphony-youri, line 15: Could we override the setter and have a String content field and a String template field (with the setter assigning the proper field?)

workflow-language/src/main/resources/swadl-schema-1.0.json
```diff
@@ -2920,6 +2920,10 @@
                         "<mention uid='${event.initiator.user.userId}'/>"
                     ]
                 },
+                {
```
symphony-youri, line 2923: can we enforce it to be template? (so we get code completion)

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/message/SendMessageExecutor.java
```diff
@@ -105,6 +111,27 @@ private Message buildMessage(ActivityExecutorContext<SendMessage> execution) thr
...
+  static String extractContent(Object contentObj, Map<String, Object> parameters, MessageService messageService)
+      throws IOException {
+    String content;
+    try {
+      String template = OBJECT_MAPPER.convertValue(contentObj, SendMessage.Content.class)
+          .getTemplate();
+      Map<String, String> params = parameters.entrySet().stream()
+          .collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue().toString()));
+      File file = new File(template);
+      if (file.getName().endsWith(".ftl")) {
```
symphony-youri, line 123: can't we consider all the files to be templates? would that break something?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/CamundaExecutor.java
```diff
@@ -147,6 +147,11 @@ public void setOutputVariable(String name, Object value) {
       execution.setVariable(String.format("%s.%s.%s", activityId, ActivityExecutorContext.OUTPUTS, name), valueVar);
     }
 
+    @Override
+    public Map<String, Object> getVariables() {
+      return (Map<String, Object>) execution.getVariable(VARIABLES);
```
symphony-youri, line 152: let's maybe return a copy of that to avoid using it as a way to change variables (not sure if it would be possible though)

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/message/SendMessageExecutor.java
```diff
@@ -105,6 +111,27 @@ private Message buildMessage(ActivityExecutorContext<SendMessage> execution) thr
...
+        content = messageService.templates().newTemplateFromFile(file.getPath()).process(params);
```
symphony-youri, line 124: to access the the file we might want to use the getResource from the context, that will get you a file from the workflows folder

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/message/SendMessageExecutor.java
symphony-mariacristina: yes, for some reason I tested it before and it was not working. But now it is, so probably I had a different error. 
We can use directly the messageService template from the bdk in both cases.

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/CamundaExecutor.java
symphony-soufiane: Assuming we use copies, what if the variable changes during workflow execution (with a script), do we want to keep the first value (that is in the copy) or the up-to-date one?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/message/SendMessageExecutor.java
symphony-soufiane: You can check how we load attachment file using that, in this executor :)

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/CamundaExecutor.java
symphony-soufiane: What about prefixing template related variables or putting them in a dedicated object so we don't mix them with regular variables. 
Like this a regular variable can be used with:  **_variables.myVariable_**, while template's ones can be used with: **_variables.templates.myVariable_**  or even **_variables.templates.SOME_ID_RELATED_TO_FILE.myVariable_**. 

Does it worth doing this or not?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/message/SendMessageExecutor.java
```diff
@@ -105,6 +111,27 @@ private Message buildMessage(ActivityExecutorContext<SendMessage> execution) thr
     return builder.build();
   }
 
+  static String extractContent(Object contentObj, Map<String, Object> parameters, MessageService messageService)
```
symphony-soufiane, line 114: Maybe we can provide TemplateEngine instead of the whole MessageService

workflow-bot-app/build.gradle
```diff
@@ -10,7 +10,7 @@ javadoc {
 dependencies {
     implementation project(':workflow-language')
 
-    implementation platform('org.finos.symphony.bdk:symphony-bdk-bom:2.4.0')
+    implementation platform('org.finos.symphony.bdk:symphony-bdk-bom:2.4.1')
```
symphony-mariacristina, line 13: we need to release a new bdk version

workflow-language/src/main/resources/swadl-schema-1.0.json
symphony-youri: if we can address that , otherwise LGTM

symphony-mariacristina: Done in the next commit

docs/reference.md
```diff
@@ -775,6 +780,23 @@ activities:
       content:
         ${text(event.source.message.message)}
 ```
+Example using freemarker:
```
symphony-youri, line 783:
```suggestion
Example using Freemarker:
```

docs/reference.md
```diff
@@ -759,6 +759,11 @@ Content can
 be [MessageML](https://docs.developers.symphony.com/building-bots-on-symphony/messages/overview-of-messageml) with
 the `<messageML>` tags or can be simple text too (<messageML> are automatically added if needed).
 
+Content can either set directly in the swadl file as plain text or it can be references from an external file. Both freemarker or mml.xml
+format are accepted as external files. By default it will search for the template file in the `./workflows` root folder.
+When freemarker is used, any workflow variable can be referenced in the external file, same format as it is for the any
```
symphony-youri, line 764:
```suggestion
When Freemarker is used, any workflow variable can be referenced in the external file, same format as it is for the any
```

symphony-youri: We can also link to https://freemarker.apache.org/

docs/reference.md
```diff
@@ -759,6 +759,11 @@ Content can
 be [MessageML](https://docs.developers.symphony.com/building-bots-on-symphony/messages/overview-of-messageml) with
 the `<messageML>` tags or can be simple text too (<messageML> are automatically added if needed).
 
+Content can either set directly in the swadl file as plain text or it can be references from an external file. Both freemarker or mml.xml
```
symphony-youri, line 762:
```suggestion
Content can either set directly in the SWADL file as plain text or it can be references from an external file. Both freemarker or mml.xml
```

docs/reference.md
```diff
@@ -759,6 +759,11 @@ Content can
...
+format are accepted as external files. By default it will search for the template file in the `./workflows` root folder.
+When freemarker is used, any workflow variable can be referenced in the external file, same format as it is for the any
+other activity in the swadl file.
```
symphony-youri, line 765:
```suggestion
other activity in the SWADL file.
```

docs/reference.md
```diff
@@ -687,7 +687,7 @@ Posts a message to a stream. Probably the most commonly used activity to interac
 Key | Type | Required |
 ------------ | -------| --- | 
 [to](#to) | Map | No |
-[content](#send-message-content) | String | Yes |
+[content](#send-message-content) | String or Object | Yes |
```
symphony-youri, line 690: Then we should probably describe what the object is in #send-message-content

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/SendMessageIntegrationTest.java
```diff
@@ -142,6 +152,30 @@ void sendMessageWithUidsVariables() throws Exception {
     assertThat(workflow).isExecuted().hasOutput(String.format(OUTPUTS_MSG_KEY, "sendMessageWithUserIds"), message);
   }
 
+  @Test
+  void sendMessageWithTemplateSuccessfull() throws IOException, ProcessingException {
+    final Workflow workflow =
+        SwadlParser.fromYaml(getClass().getResourceAsStream("/message/send-message-with-freemarker.swadl.yaml"));
+
+    TemplateEngine templateEngine = mock(TemplateEngine.class);
```
symphony-youri, line 160: Could we have used the real template engine here?

symphony-youri: Otherwise are we really testing the variable binding?

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/SendMessageIntegrationTest.java
```diff
@@ -142,6 +152,30 @@ void sendMessageWithUidsVariables() throws Exception {
...
+    engine.onEvent(messageReceived("/send"));
+
+    assertThat(workflow).isExecuted();
+    ArgumentCaptor<Message> messageArgumentCaptor = ArgumentCaptor.forClass(Message.class);
+    ArgumentCaptor<Map<String, Object>> paramsArgumentCaptor = ArgumentCaptor.forClass(Map.class);
+
+    verify(template, times(1)).process(paramsArgumentCaptor.capture());
+    assertThat(paramsArgumentCaptor.getValue()).contains(Map.entry("variables",  Map.of("val","world")));
+    verify(messageService, times(1)).send(eq("123"), messageArgumentCaptor.capture());
+    AssertionsForClassTypes.assertThat(messageArgumentCaptor.getValue().getContent()).isEqualTo("<messageML>Hello world!</messageML>");
```
symphony-youri, line 176: is that the proper assertThat to import? (vs the one above)

symphony-mariacristina: removed the import in the next commit

docs/reference.md
symphony-mariacristina: I improved the doc in the last commit. What do you think?

docs/reference.md
```diff
@@ -759,6 +759,18 @@ Content can
 be [MessageML](https://docs.developers.symphony.com/building-bots-on-symphony/messages/overview-of-messageml) with
 the `<messageML>` tags or can be simple text too (<messageML> are automatically added if needed).
 
+Content can either set directly in the SWADL file as plain text (String) or it can be referenced from an external file (Object).
```
symphony-soufiane, line 762: [...] can "be" set [...]

symphony-youri (approved):

PR #65, opened 2021-11-18
PLAT-11535: Timeout support without using subprocess

symphony-soufiane:
Related to [PLAT-11535](https://perzoinc.atlassian.net/browse/PLAT-11535)

### Description
To support timeout in different activities, an implementation was done based on subprocesses but it turned out that it is not the correct implementation in terms of flow control as subprocesses can be rexecuted multiple times.
In this commit, we implement it in the main process. For this occasion, we introduced EventWithTimeout than represents an event that can have a timeout, like this in the java model we do not allow to have timeouts in one-of event and it can only be set at (on) property level.

The workflow BPMN we wanted to be able to build with an activity being a kind of (try/catch) is executed when 1 or more activities expired. (built with this PR code)

bpmn


workflow-language/src/main/resources/swadl-schema-1.0.json
```diff
@@ -666,7 +666,8 @@
             },
             "required": [
                 "message-received"
-            ]
+            ],
+            "additionalProperties": true
```
symphony-youri, line 670: do we have to change all events?

symphony-soufiane: It looks like this is not needed since we don't consider timeout if defined inside the event. Adding it for "one-of" property is enough. I'll remove it in next commit.

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/custom/assertion/WorkflowAssert.java
```diff
@@ -29,6 +29,10 @@
 import java.util.stream.Collectors;
 
 public class WorkflowAssert extends AbstractAssert<WorkflowAssert, Workflow> {
+
+  private static final List<String> activityTypesToIgnore =
```
symphony-youri, line 33: ACTIVITY_TYPES_TO_IGNORE since it is a constant

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/EventTypesIntegrationTest.java
```diff
@@ -131,6 +133,51 @@ void onMessageReceived_timeout(String swadlFile, String messageToReceive)
...
+  void onActivityExpired_timeout() throws IOException, ProcessingException, InterruptedException {
+    final Workflow workflow =
+        SwadlParser.fromYaml(getClass().getResourceAsStream("/event/timeout/activity-expired-with-timeout.swadl.yaml"));
+
+    when(messageService.send(anyString(), any(Message.class))).thenReturn(new V4Message());
+
+    engine.deploy(workflow);
+    engine.onEvent(messageReceived("/start"));
+
+    Thread.sleep(500); // wait 0.5s to let workflow times out
```
symphony-youri, line 146: if we start repeating it let's make it a method (and its name will replace the comment)

symphony-youri (approved):

PR #66, opened 2021-11-22
PLAT-11536 Fixed multipart content type.

symphony-soufiane:
When sending a multipart/form request, the content type should include the boundary. However, this is no something the user can set in SWADL.
In this commit, we changed the way we handle this content type request and we made use of MultipartEntityBuilder that sets the boundary along with the content type in body.
When the content type is provided in the header than it overrides the one set by MultipartEntityBuilder. For this reason, when we process a multipart/form request, we remove the content type from headers if the user has provided any in order to make use of the one set with boundary.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1874,7 +1874,8 @@ Key | Type | Required |
 [headers](#headers) | String | No |
 
 _nb: For multipart/form-data content type requests, the body should be provided as a key/value object. For other content types, it can be provided as String in JSON format._
-_nb: For a JSON api, the response can be processes in subsequent activities in SWADL using the utility function [json()](#utility-functions)
+
+_nb: For a JSON api, the response can be processed in subsequent activities in SWADL using the utility function [json()](#utility-functions)_
 
 Output | Type |
 ----|----|
@@ -1964,7 +1965,7 @@ activities:
 Script to execute (only [Groovy](https://groovy-lang.org/) is supported).
 
 ## Utility functions
-WDK provides some utility functions that can be used in SWADL to process data in SWADL.
+WDK provides some utility functions that can be used to process data in SWADL.
 
 ### Object json(String string)
 This method is used to convert a String in JSON format to an Object in order to be processed as a JSON.
@@ -1977,7 +1978,7 @@ ${json("This is a regular String")} == "This is a regular String"
 ```
 
 ### String text(String presentationMl)
-This method is used to convert a presentation ML String to a text.
+This method is used to convert a PresentationML String to a text.
 
 Example:
 ```java
@@ -1985,7 +1986,7 @@ ${text(<div data-format="PresentationML" data-version="2.0">started</div>)} == "
 ```
 
 ### String escape(String string)
-This method will escape text contents using JSON standard escaping, and return results as a String
+This method will escape text contents using JSON standard escaping, and return results as a String.
 
 Example:
 ```java
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1874,6 +1874,7 @@ Key | Type | Required |
 [headers](#headers) | String | No |
 
 _nb: For multipart/form-data content type requests, the body should be provided as a key/value object. For other content types, it can be provided as String in JSON format._
+_nb: For a JSON api, the response can be processes in subsequent activities in SWADL using the utility function [json()](#utility-functions)
 
 Output | Type |
 ----|----|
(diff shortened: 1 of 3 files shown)


workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/client/HttpClient.java
```diff
@@ -18,6 +19,7 @@
 
 @Generated
 @Component
+@Slf4j
 public class HttpClient {
 
   public Response execute(String method, String url, Object body, Map<String, String> headers)
```
symphony-youri, line 25: as it is getting more complex maybe we should invest in wiremock or something to help us test this activity

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/RequestExecutor.java
```diff
@@ -29,11 +31,13 @@ public RequestExecutor(HttpClient httpClient) {
   @Override
   public void execute(ActivityExecutorContext<ExecuteRequest> execution) throws IOException {
     ExecuteRequest activity = execution.getActivity();
+    log.info("Executing request {} {}", activity.getMethod(), activity.getUrl());
 
     Response response =
         this.httpClient.execute(activity.getMethod(), activity.getUrl(), activity.getBody(),
             headersToString(activity.getHeaders()));
 
+    log.info("Received response {}", response.getContent());
```
symphony-youri, line 40: it might contain sensitive data (and be very big) so we might want to only log this in debug (or not log it at all)

docs/reference.md
```diff
@@ -1874,6 +1874,7 @@ Key | Type | Required |
 [headers](#headers) | String | No |
 
 _nb: For multipart/form-data content type requests, the body should be provided as a key/value object. For other content types, it can be provided as String in JSON format._
+_nb: For a JSON api, the response can be processes in subsequent activities in SWADL using the utility function [json()](#utility-functions)
```
symphony-youri, line 1877: processes -> processed

docs/reference.md
```diff
@@ -1961,3 +1962,32 @@ activities:
 #### script
 
 Script to execute (only [Groovy](https://groovy-lang.org/) is supported).
+
+## Utility functions
+WDK provides some utility functions that can be used in SWADL to process data in SWADL.
```
symphony-youri, line 1967:
```suggestion
WDK provides some utility functions that can be used to process data in SWADL.
```

docs/reference.md
```diff
@@ -1961,3 +1962,32 @@ activities:
...
+It returns a String if the parameter is a simple String.
+
+Example:
+```java
+${json("{\"result\": {\"code\": 200, \"message\": \"success\"}}").result.code} == 200
+${json("This is a regular String")} == "This is a regular String"
+```
+
+### String text(String presentationMl)
+This method is used to convert a presentation ML String to a text.
```
symphony-youri, line 1980:
```suggestion
This method is used to convert a PresentationML String to a text.
```

docs/reference.md
```diff
@@ -1961,3 +1962,32 @@ activities:
...
+
+Example:
+```java
+${text(<div data-format="PresentationML" data-version="2.0">started</div>)} == "started"
+```
+
+### String escape(String string)
+This method will escape text contents using JSON standard escaping, and return results as a String
```
symphony-youri, line 1988:
```suggestion
This method will escape text contents using JSON standard escaping, and return results as a String.
```

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/client/HttpClient.java
```diff
@@ -18,6 +19,7 @@
 
 @Generated
 @Component
+@Slf4j
```
symphony-youri, line 22: is it used?

symphony-youri (commented):
Thanks for the doc on utility functions

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/RequestExecutor.java
symphony-soufiane: What about only logging the status code? the idea here is to see if the request has been fulfilled and if it succeeded or not, because we might not use the outputs in SWADL

symphony-youri (approved):

PR #70, opened 2021-11-25
PLAT-11535: Support timeout.

symphony-soufiane:
[PLAT-11535](https://perzoinc.atlassian.net/browse/PLAT-11535)
In a previous PR we added the support of timeouts with the main process and we added the possibility to connect the timed out activities to a dedicated activity. This latter is executed when the specified one/ones is/are timed out. However, this implementation did not allow to connect a branch (a part of workflow) to activities with time out. Only one acitivty could be connected.
The issue was that when we have an activity with on:acitivty-expired property, we move the builder backward to the specified activity having timeout property in order to connect the two activities, but we do not update the builder with the new one. Thus, the following activities are added to the end of the workflow.
With this implementation, having an activity ac1 with on:activity-expired property, if we define an activity ac2 just after it, then ac1 will be connected to ac2. In order to connect ac2 to the end of the workflow instead of the new branch, it needs to use on:activity-completed referring the activity with timeout.

to do:
- [] Update reference documentation
- [] Add examples for different cases
----
**Here is an example of SWADL with the generated BPMN**
Screenshot 2021-11-25 at 20 40 02

----
_By adding an activity to the main branch of workflow_:
x

### DIFF
--- a/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/bpmn/CamundaBpmnBuilder.java
+++ b/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/bpmn/CamundaBpmnBuilder.java
@@ -134,11 +134,9 @@ public class CamundaBpmnBuilder {
     for (Activity activityContainer : workflow.getActivities()) {
       BaseActivity activity = activityContainer.getActivity();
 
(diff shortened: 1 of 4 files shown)


symphony-youri (approved):
👍 clear explanations with the PR

symphony-mariacristina (approved):

PR #71, opened 2021-11-29
Improvement of execute-request activity

symphony-soufiane:
Execute-request activity supports now a body that can be either put in json string or object when the content type is application/json. HttpClient is able to do the required conversion depending on the body instance type.
This change makes the workflow writing easier.
Body field in Response object is now an Object to support both string and Map.

- Example of POST api with application/json content type: https://github.com/finos/symphony-wdk/pull/71/files#diff-fc9d7bd0a5454a824203bfedd1178616effb2ff0dcc945734024529f51eadbbf
- Example of GET api with application/json content type: https://github.com/finos/symphony-wdk/pull/71/files#diff-835c4a40b5a09383b00b2f78fed7ba4ae6fccf43a06f1abb820ad52a0dae97a1
- Example of POST api with multipart content type: https://github.com/finos/symphony-wdk/pull/71/files#diff-9142527260268ce491f2b305ae1d822bb08374b759c24e5f8a492688d10070c6


docs/examples/call-create-message-multipart-api.swadl.yaml
```diff
@@ -0,0 +1,20 @@
+id: create-message-request-workflow
+activities:
+  - execute-request:
```
symphony-youri, line 3: I'm not sure having an example for an API call that is covered by an existing activity is good
It could lead user to think that this is how they can create a msg

docs/examples/call-create-room-json-api.swadl.yaml
```diff
@@ -0,0 +1,18 @@
+id: create-room-request-workflow
+activities:
```
symphony-youri, line 2: isn't the example with the exchange rate api enough in the end?

docs/reference.md
```diff
@@ -1873,9 +1873,9 @@ Key | Type | Required |
 [body](#body) | Object/String | No |
 [headers](#headers) | String | No |
 
-_nb: For multipart/form-data content type requests, the body should be provided as a key/value object. For other content types, it can be provided as String in JSON format._
+_nb: For multipart/form-data content type requests, the body should be provided as a key/value object. For other content types, String in JSON format is also supported._
 
-_nb: For a JSON api, the response can be processed in subsequent activities in SWADL using the utility function [json()](#utility-functions)_
+_nb: If the api response is not JSON, then it can be processed in subsequent activities in SWADL using the utility function [json()](#utility-functions)_
```
symphony-youri, line 1878: if it not JSON content while would you use the json() function?

docs/reference.md
```diff
@@ -1930,7 +1916,7 @@ Supported methods are:
 - PUT
 
 #### body
-HTTP request body. Depending on the Content-Type, the body can be an Object when _multipart/form-data_ is used or a String for the other content types.   
+HTTP request body. Depending on the Content-Type, the body can be an Object when _multipart/form-data_ is used. String is also support for the other content types.
```
symphony-youri, line 1919: supported

docs/reference.md
```diff
@@ -1873,9 +1873,9 @@ Key | Type | Required |
 [body](#body) | Object/String | No |
 [headers](#headers) | String | No |
 
-_nb: For multipart/form-data content type requests, the body should be provided as a key/value object. For other content types, it can be provided as String in JSON format._
+_nb: For multipart/form-data content type requests, the body should be provided as a key/value object. For other content types, String in JSON format is also supported._
```
symphony-youri, line 1876: if we support JSON now why don't we say so?

docs/reference.md
```diff
@@ -1930,7 +1916,7 @@ Supported methods are:
 - PUT
 
 #### body
-HTTP request body. Depending on the Content-Type, the body can be an Object when _multipart/form-data_ is used or a String for the other content types.   
+HTTP request body. Depending on the Content-Type, the body can be an Object when _multipart/form-data_ is used. String is also support for the other content types.   
 When the body is provided, Content-Type must be set in headers.
```
symphony-youri, line 1920: it is JSON by default

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/client/HttpClient.java
```diff
@@ -29,13 +34,28 @@ public Response execute(String method, String url, Object body, Map<String, Stri
...
+          IOUtils.toString(classicHttpResponse.getEntity().getContent(), StandardCharsets.UTF_8),
+          classicHttpResponse.getFirstHeader(HttpHeaders.CONTENT_TYPE));
     } else {
       return new Response(responseCode, "");
     }
   }
 
+  private Response handleResponse(int statusCode, String content, Header contentType) {
+    Object data = content;
+    if (contentType == null || (contentType.getName().equals(HttpHeaders.CONTENT_TYPE) && contentType.getValue()
```
symphony-youri, line 47: we can introduce a isJsonContent() private method or something to help read the complex condition

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/client/HttpClient.java
```diff
@@ -29,13 +34,28 @@ public Response execute(String method, String url, Object body, Map<String, Stri
     int responseCode = httpResponse.getCode();
 
     if (classicHttpResponse.getEntity() != null && classicHttpResponse.getEntity().getContent() != null) {
-      return new Response(responseCode,
-          IOUtils.toString(classicHttpResponse.getEntity().getContent(), StandardCharsets.UTF_8));
+      return this.handleResponse(responseCode,
+          IOUtils.toString(classicHttpResponse.getEntity().getContent(), StandardCharsets.UTF_8),
+          classicHttpResponse.getFirstHeader(HttpHeaders.CONTENT_TYPE));
```
symphony-youri, line 39: should not we pass the value directly (to avoid checking the header's name again in handleResponse)?

symphony-youri (commented):
As for the failing sendMessageOnMessage test maybe it comes from the fact that we mock twice the     when(messageService.send(anyString(), any(Message.class))).thenReturn(message);

and mockito uses one of the other mock

docs/examples/call-create-message-multipart-api.swadl.yaml
symphony-soufiane: I agree this will lead to confusion. I'm removing them in the next commit.

docs/examples/call-create-room-json-api.swadl.yaml
symphony-soufiane: exchange rate api is a GET and does not show usage of the request body. But we can keep this one.

docs/reference.md
symphony-soufiane: I'm removing json() method in next commit. it is not needed anymore

docs/reference.md
symphony-soufiane: Because JSON string is not supported for multipart/form-data. 
I updated this in "body" description to say that it can be provided as JSON or key/value object, but only key/value when using mulitpart/form.

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/IntegrationTest.java
```diff
@@ -122,7 +120,7 @@ protected static Stream stream(String streamId) {
 
   @BeforeEach
   void setUpMocks() {
-    when(messageService.send(anyString(), any(Message.class))).thenReturn(message("msgId"));
+    //when(messageService.send(anyString(), any(Message.class))).thenReturn(message("msgId"));
```
symphony-youri, line 123: that's commented code, please remove it

symphony-youri (approved):

PR #74, opened 2021-12-02
Support loop after form replies

symphony-youri:
The main problem is that form replies start subprocesses each one having
its own life. But sometime in 1-1 or in specific case a form is meant to
be replied by only one user. Meaning we can avoid using the subprocess
and allow for loops to happen after the form is being replied to.

Introduce a flag in the form-reply event to decide wether subprocesses
are used or not. The timeouts are not yet handled.


symphony-soufiane:
Don't forget to change https://github.com/finos/symphony-wdk/blob/master/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/CamundaExecutor.java#L235

PR #75, opened 2021-12-02
Update Readme

symphony-youri:
Don't maintain list of issues manually, just link to the issues in Github


symphony-soufiane:
The idea of putting that list was to give a slight idea of the key features we will have in a near future. But I agree this requires us to maintain it which we can avoid by linking to Github issues.

symphony-soufiane (approved):

PR #76, opened 2021-12-02
Fix flaky test sendMessageOnMessage

symphony-youri:
PR checks are sometimes failing. I'm not 100% sure but maybe even if the
process is marked as completed all the historic details (like vars) have
not yet been written to the DB.

So we wait (actively) for that when checking variables.

### DIFF
--- a/workflow-bot-app/src/test/java/com/symphony/bdk/workflow/custom/assertion/WorkflowAssert.java
+++ b/workflow-bot-app/src/test/java/com/symphony/bdk/workflow/custom/assertion/WorkflowAssert.java
@@ -205,15 +205,15 @@ public class WorkflowAssert extends AbstractAssert<WorkflowAssert, Workflow> {
     String process = lastProcess().orElseThrow();
     await().atMost(5, SECONDS).until(() -> processIsCompleted(process));
 
-    await().atMost(5, SECONDS).until(()
-        -> !IntegrationTest.historyService.createHistoricDetailQuery().processInstanceId(process).list().isEmpty());
-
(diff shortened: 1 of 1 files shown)


symphony-soufiane (approved):
This will not make the tests slower? But if it fixes the flaky tests, let's do it.

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/custom/assertion/WorkflowAssert.java
```diff
@@ -205,12 +205,15 @@ private void assertOutputs(String key, Object value) {
     String process = lastProcess().orElseThrow();
     await().atMost(5, SECONDS).until(() -> processIsCompleted(process));
 
-    final List<HistoricDetail> historicalDetails =
-        IntegrationTest.historyService.createHistoricDetailQuery().processInstanceId(process).list();
-
-    Optional<HistoricDetail> historicalDetailOptional = historicalDetails.stream()
-        .filter(x -> ((HistoricDetailVariableInstanceUpdateEntity) x).getVariableName().equals(key))
-        .findFirst();
+    Optional<HistoricDetail> historicalDetailOptional = await().atMost(5, SECONDS).until(()
```
symphony-soufiane, line 208: Do we need a similar fix in https://github.com/finos/symphony-wdk/pull/76/files#diff- 47eda30bc6637c1e8de6cf0b8f4546dbc96c706f202559e36c897e789cf3181dL155 ?

symphony-youri: It looks like so far we did not have issues with this one, let's keep it in mind if we have flaky tests again

PR #84, opened 2021-12-07
PLAT-11620: Support looping on single-replied form

symphony-soufiane:
The main problem is that form replies start subprocesses each one having its own life. But sometime in 1-1 or in specific case, a form is meant to be replied by only one user. Meaning we can avoid using the subprocess and allow for loops to happen after the form is being replied.

In this PR, we Introduced a flag in the form-reply event to decide wether subprocesses are used or not.
Timeout is implemented in the same way as in the other activities. If no timeout is provided, then the default PT24H is used.

In terms of implementation, the timeout for single reply forms is done in the same method as for the other activities. This allows us to maintain one method for both of them and we ensure having the same behaviour.

**_Example of a workflow having timeout on a single reply form with subsequent activities:_**
workflow-that-has-a-flow

### DIFF
--- a/docs/examples/approval-form.swadl.yaml
+++ b/docs/examples/approval-form.swadl.yaml
@@ -1,0 +1,56 @@
+id: approval-form
+activities:
+  - send-message:
+      id: sendForm
+      on:
+        one-of:
+          - message-received:
+              content: /send
+          - activity-completed:
+              activity-id: setVariables
+              if: ${variables.loop == true}
+      content: |
+        <messageML>
+          <p>Do you approve the action?</p>
+          <form id="sendForm">
+            <button name="approve" type="action">Approve</button>
+            <button name="reject" type="action">Reject</button>
+          </form>
+        </messageML>
+
+
+  # Form requiring only one reply from the admin
+  # 10s to reply before it expires
+  - execute-script:
+      id: setVariables
+      on:
+        timeout: PT10S
+        form-replied:
+          form-id: sendForm
+          unique: true
+      script: |
+        variables.loop = (sendForm.action == "reject" )
+
+  - send-message:
+      id: sendFeedBackOnReply
+      content: The form has been approved!
+
+  - send-message:
+      id: sendFeedBackOnExpiration
+      content: The form timed out, the workflow will finish immediately
+      on:
+        activity-expired:
+          activity-id: setVariables
+
+  - send-message:
+      id: finishWorkflowOnExpiration
+      content:  The workflow finished due to form timeout
+
+  - send-message:
+      id: finishWorkflowOnReply
+      content: The workflow finished after having replied to the form
+      on:
+        activity-completed:
+          activity-id: sendFeedBackOnReply
+
+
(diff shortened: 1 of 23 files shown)


workflow-bot-app/src/test/resources/form/send-form-unique-reply-timeout.swadl.yaml
```diff
@@ -0,0 +1,30 @@
...
+      to:
+        stream-id: "123"
+
+  - execute-script:
+      id: replyWithTimeout
+      on:
+        timeout: PT0.01S
+        form-replied:
+          form-id: sendForm
+          unique: true
```
symphony-youri, line 24: we should figure out what is the best name for it, unique, reply ? and what should be the default value
it is probably worth discussing with dev rel team

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/CamundaExecutor.java
```diff
@@ -20,6 +18,7 @@
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.json.JsonMapper;
 import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.google.common.collect.ImmutableMap;
```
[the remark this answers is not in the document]
symphony-soufiane: removed in last commit

workflow-bot-app/src/test/resources/form/send-form-unique-reply-timeout.swadl.yaml
symphony-soufiane: I keep and document this waiting for feedback from devRel team

docs/examples/approval-form.swadl.yaml
```diff
@@ -0,0 +1,56 @@
...
+        variables.loop = (sendForm.action == "reject" )
+
+  - send-message:
+      id: sendFeedBackOnReply
+      content: The form has been approved!
+
+  - send-message:
+      id: sendFeedBackOnExpiration
+      content: The form timed out, the workflow will finish immediately
+      on:
```
symphony-youri, line 41: I personally prefer to have the on/event right after the id (and also to be consistent with the activity above)

docs/reference.md
```diff
@@ -263,9 +263,15 @@ As multiple users can reply to a form sent to a room, the activity waiting for a
 times (it can be seen as a sub execution of the current workflow). This also means that every activity defined after
 this activity or with a `activity-completed` event on this activity will run within its own sub execution.
 
+A form can be dedicated to be replied only once. In this case, `unique` is set to `true`.
```
symphony-youri, line 266: In 1-1 conversations or if the form has to be replied only once, the unique attribute on the form-replied event can be set to true. In that case the activity in the main execution of the current workflow and flow controls can be used as normal.

symphony-youri:
`exclusive` was suggested too instead of unique, I don't have a strong opinion on this, in any case it will require users to take a look at the doc

symphony-youri (approved):

PR #85, opened 2021-12-07
PLAT-11617: Expose util functions in execute-script activity

symphony-soufiane:
UtilFunctionsMapper provides utility function that can be used in SWADL. So far, it was required to import them or to use to full path in order to call a utility method.
In this method, we created a process engine plugin that injects UtilFunctionsMapper in Camunda script engine. We used utilityFunctionsMapper as a key in order to call method with utilityFunctionsMapper.escape(..) but this key can be changed to a better one if needed.
Doc and tests are updated following this change.


symphony-youri (commented):
Have you tried to find a way to do it with accessing the functions directly in the script?

(I could not find a good one when I tried, and also I'm not sure it is bad to have to prefix the function call as a way to "namespace" it and avoid conflicts)

symphony-youri (approved):

PR #96, opened 2022-03-21
PLAT-11870: Prepare GA release

symphony-soufiane:
### Description
Added a log of wdk version using Gradle resource filtering. A placehoder is added in spring application.yaml file that is replaced after the build with the current version.
Removed a disclaimer from README.

### Dependencies
N/A

### Checklist

### DIFF
--- a/workflow-bot-app/build.gradle
+++ b/workflow-bot-app/build.gradle
@@ -1,3 +1,5 @@
+import org.apache.tools.ant.filters.ReplaceTokens
+
 plugins {
     id 'workflow-bot.java-conventions'
     id 'org.springframework.boot' version '2.6.3'
@@ -81,4 +83,10 @@ publishing {
             artifact bootJar
         }
     }
+}
+
+processResources {
+    filter ReplaceTokens, tokens: [
+            "wdk.version": project.version
+    ]
 }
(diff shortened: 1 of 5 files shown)


symphony-youri:
I'd we usually don't reference our internal ticker numbers in OSS projects as they don't mean much for external contributors.
What about using https://docs.spring.io/spring-boot/docs/current/gradle-plugin/reference/htmlsingle/#integrating-with-actuator.build-info  and then injecting https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/info/BuildProperties.html?

workflow-bot-app/build.gradle
```diff
@@ -82,3 +84,9 @@ publishing {
         }
     }
 }
+
+processResources {
```
symphony-youri, line 88: actually


processResources {
    filesMatching('application.yaml') {
        expand(project.properties)
    }
}

could be used to also set:


bdk:
  default-headers:
    User-Agent: "Symphony-WDK/${project.version} Java/\${java.version}"

in the default application yaml

symphony-youri: our yours is simpler to make it work, no need to escape

symphony-soufiane:
ok to test

symphony-youri (approved):

PR #107, opened 2022-05-04
Added encodeRawQuery and decodeRawQuery to encode/decode url query parameters

### LINKED ISSUE #106: Add a utility method to encode/decode url parameters
### Description of Problem:
When using variables for query param in execute-request activity, they are not encoded. Given that the execute-request takes the full URL as an input and does not handle them in separate inputs, it would be the responsibility of the workflow to encode/decode the data. For that, it would be helpful to have a utility method to be called in execute-script activities.

### Potential Solutions:
Have a utility method to be called in execute-script activities.

symphony-soufiane:
### Description
Since the execute-request activity takes the full URL as input and does not handle query parameters separately, the encoding/decoding should be the responsibility of the workflow and not the activity itself. The utility methods are added to be used in execute-script.
encodeRawQuery() only encodes the query parameters and does not change the other components.

closes #106 

### Checklist

### DIFF
--- a/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/UtilityFunctionsMapper.java
+++ b/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/UtilityFunctionsMapper.java
@@ -12,12 +12,20 @@ import com.fasterxml.jackson.core.io.JsonStringEncoder;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.camunda.bpm.engine.impl.javax.el.FunctionMapper;
 import org.camunda.bpm.engine.impl.util.ReflectUtil;
+import org.springframework.util.MultiValueMap;
+import org.springframework.web.util.UriComponents;
+import org.springframework.web.util.UriComponentsBuilder;
 
+import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Method;
+import java.net.URLDecoder;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * Utilities for EL evaluation by Camunda.
@@ -65,6 +73,47 @@ public class UtilityFunctionsMapper extends FunctionMapper {
       return null;
     }
     return new String(JsonStringEncoder.getInstance().quoteAsString(s));
+  }
+
+  public static String encodeRawQuery(String url) throws UnsupportedEncodingException {
+    return encodeRawQuery(url, StandardCharsets.UTF_8.name());
+  }
+
+  public static String encodeRawQuery(String url, String encoding)
+      throws UnsupportedEncodingException {
+    UriComponents uriComponents = UriComponentsBuilder.fromUriString(url).build();
+    MultiValueMap<String, String> queryParamsMap = uriComponents.getQueryParams();
+
+    String rawQuery = uriComponents.getQueryParams()
+        .keySet()
+        .stream()
+        .map(key ->
+            key + "=" + queryParamsMap.get(key).get(0))
+        .collect(Collectors.joining("&"));
+
+    rawQuery = URLEncoder.encode(rawQuery, encoding);
+
+    StringBuilder encodedBuilder = new StringBuilder(uriComponents.getScheme() + "://" + uriComponents.getHost());
+
+    if (uriComponents.getPort() != -1) {
+      encodedBuilder.append(":").append(uriComponents.getPort());
+    }
+
+    if (uriComponents.getPath() != null) {
+      encodedBuilder.append(uriComponents.getPath()).append("?");
+    }
+
+    encodedBuilder.append(rawQuery);
+
+    return encodedBuilder.toString();
+  }
+
+  public static String decodeRawQuery(String url) throws UnsupportedEncodingException {
+    return URLDecoder.decode(url, StandardCharsets.UTF_8.name());
+  }
+
+  public static String decodeRawQuery(String url, String encoding) throws UnsupportedEncodingException {
+    return URLDecoder.decode(url, encoding);
   }
 
   @SuppressWarnings("rawtypes")
--- a/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/UtilityFunctionsMapper.java
+++ b/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/UtilityFunctionsMapper.java
@@ -10,23 +10,14 @@ import com.symphony.bdk.workflow.engine.executor.EventHolder;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.io.JsonStringEncoder;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import org.apache.commons.lang3.StringUtils;
 import org.camunda.bpm.engine.impl.javax.el.FunctionMapper;
 import org.camunda.bpm.engine.impl.util.ReflectUtil;
-import org.springframework.util.MultiValueMap;
-import org.springframework.web.util.UriComponents;
-import org.springframework.web.util.UriComponentsBuilder;
 
-import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Method;
-import java.net.URLDecoder;
-import java.net.URLEncoder;
-import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
 
 /**
  * Utilities for EL evaluation by Camunda.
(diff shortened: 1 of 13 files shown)


workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/UtilityFunctionsMapper.java
```diff
@@ -67,6 +76,49 @@ public static String escape(String s) {
...
+    String rawQuery = uriComponents.getQueryParams()
+        .keySet()
+        .stream()
+        .map(key ->
+            key + "=" + queryParamsMap.get(key).get(0))
+        .collect(Collectors.joining("&"));
+
+    rawQuery = URLEncoder.encode(rawQuery, encoding);
+
+    StringBuilder encodedBuilder = new StringBuilder(uriComponents.getScheme() + "://" + uriComponents.getHost());
```
symphony-youri, line 97: aren't there more components inside an URL? that we could loose here? like basic auth for instance

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/UtilityFunctionsMapper.java
```diff
@@ -67,6 +76,49 @@ public static String escape(String s) {
...
+    if (StringUtils.isNotBlank(rawQuery)) {
+      encodedBuilder.append("?");
+    }
+
+    encodedBuilder.append(rawQuery);
+
+    return encodedBuilder.toString();
+  }
+
+  public static String decodeRawQuery(String url) throws UnsupportedEncodingException {
```
symphony-youri, line 114: what would be the use case for decoding?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/UtilityFunctionsMapper.java
```diff
@@ -67,6 +76,49 @@ public static String escape(String s) {
     return new String(JsonStringEncoder.getInstance().quoteAsString(s));
   }
 
+  public static String encodeRawQuery(String url) throws UnsupportedEncodingException {
+    return encodeRawQuery(url, StandardCharsets.UTF_8.name());
+  }
+
+  public static String encodeRawQuery(String url, String encoding)
```
symphony-youri, line 83: do we want to expose that? given all workflows are probably expected to be utf-8 ?

symphony-youri (commented):
Can we avoid using a script activity for that?

I would suggest gathering use cases with real workflows to see how best the devx can be

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/ExecuteRequestExecutor.java
```diff
@@ -45,6 +52,30 @@ public void execute(ActivityExecutorContext<ExecuteRequest> execution) throws IO
...
+  private String encodeQueryParameters(String fullUrl) {
+    // splits by the first occurrence of "?" to get the query params part
+    String[] splitUrl = fullUrl.split("\\?", 2);
+    String encodedUrl = splitUrl[0];
+
+    MultiValueMap<String, String> queryParamsMap =
+        UriComponentsBuilder.fromUriString(fullUrl).build().getQueryParams();
+
+    if (!queryParamsMap.isEmpty()) {
+      log.info("Encoding query parameters with Standard UTF-8 format");
```
symphony-youri, line 64: I would put this a debug (and even then, I'm not sure I would log that)

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/ExecuteRequestExecutor.java
```diff
@@ -45,6 +52,30 @@ public void execute(ActivityExecutorContext<ExecuteRequest> execution) throws IO
...
+
+      String rawQuery = queryParamsMap
+          .keySet()
+          .stream()
+          .map(key ->
+              key + "=" + URLEncoder.encode(queryParamsMap.get(key).get(0), StandardCharsets.UTF_8))
```
symphony-youri, line 70: so if we have multiple query params in the map, what becomes of the other ones?

if the query param is already encoded, what will it do?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/ExecuteRequestExecutor.java
```diff
@@ -45,6 +52,30 @@ public void execute(ActivityExecutorContext<ExecuteRequest> execution) throws IO
     execution.setOutputVariables(outputs);
   }
 
+  private String encodeQueryParameters(String fullUrl) {
+    // splits by the first occurrence of "?" to get the query params part
+    String[] splitUrl = fullUrl.split("\\?", 2);
+    String encodedUrl = splitUrl[0];
+
+    MultiValueMap<String, String> queryParamsMap =
+        UriComponentsBuilder.fromUriString(fullUrl).build().getQueryParams();
```
symphony-youri, line 61: then cannot we keep using this class, parse the url, clear the query params and set the ones that we encoded
to avoid parsing/building the url manually

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/ExecuteRequestExecutor.java
```diff
@@ -45,6 +52,30 @@ public void execute(ActivityExecutorContext<ExecuteRequest> execution) throws IO
     execution.setOutputVariables(outputs);
   }
 
+  private String encodeQueryParameters(String fullUrl) {
```
symphony-youri, line 55: we may want to extract this and unit test it separately to cover different use cases

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/ExecuteRequestExecutor.java
symphony-elias: You could use https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/util/UriComponentsBuilder.html#replaceQueryParams-org.springframework.util.MultiValueMap-

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/UtilityFunctionsMapper.java
```diff
@@ -67,6 +76,41 @@ public static String escape(String s) {
...
+
+    return uriComponentsBuilder.build().toUriString();
+  }
+
+  private static boolean isAlreadyEncoded(String value) {
+    boolean isAlreadyEncoded = false;
+    try {
+      isAlreadyEncoded = !URLDecoder.decode(value, StandardCharsets.UTF_8.name()).equals(value);
+    } catch (UnsupportedEncodingException | IllegalArgumentException e) {
+      e.printStackTrace();
```
symphony-youri, line 108: printstacktrace goes to the console so it will not be properly handled by the logging configuration

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/UtilityFunctionsMapper.java
```diff
@@ -67,6 +76,41 @@ public static String escape(String s) {
...
+        .keySet()
+        .stream()
+        .filter(k -> !isAlreadyEncoded(queryParamsMap.get(k).get(0))).collect(Collectors.toList())
+        .forEach(
+            key -> {
+              try {
+                uriComponentsBuilder.replaceQueryParam(key,
+                    URLEncoder.encode(queryParamsMap.get(key).get(0), StandardCharsets.UTF_8.name()));
+              } catch (UnsupportedEncodingException e) {
+                e.printStackTrace();
```
symphony-youri, line 96: same here, to remove

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/utils/UtilityFunctionsMapperTest.java
```diff
@@ -49,4 +51,47 @@ void jsonEmptyTest() {
     assertThat(json.toString()).isEmpty();
   }
 
+  @Test
+  void encodeQueryParametersTest() {
```
symphony-youri, line 55: could you add a test with

https://www.wdk.symphony.com:8080/path1/path2?key1=value1&key1=value2 as the URL?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/ExecuteRequestExecutor.java
symphony-soufiane: The query params can be cleaned by setting them to Null, but we can only access them after building the UriComponents, the builder UriComponentsBuilder does not provide getter methods. 
In my last commit, I change the instantiation (fromUriString(fullUrl)) and I rather clone the builder, but I don't really see an interest of doing that.

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/ExecuteRequestUtils.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ExecuteRequestUtils {
+
+  ExecuteRequestUtils() {}
```
symphony-youri, line 14: it can be private

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/utils/ExecuteRequestUtilsTest.java
```diff
@@ -0,0 +1,37 @@
+package com.symphony.bdk.workflow.utils;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.symphony.bdk.workflow.engine.executor.request.ExecuteRequestUtils;
+
+import org.junit.jupiter.api.Test;
+
```
symphony-youri, line 8: should we have a test for parameters that are already encoded?

symphony-youri (commented):
As we discussed a test with a workflow using the encode-query-params attribute (false and true) would be interesting too

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/utils/ExecuteRequestUtilsTest.java
symphony-soufiane: Added in next commit. We decided, in this case, to re-encode the query params without any check. If the workflow developer wants the url to be automatically encoded, he should not provide an encoded one. Otherwise, he can disable it using the optional field.

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/ExecuteRequestUtils.java
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.util.MultiValueMap;
+import org.springframework.web.util.UriComponentsBuilder;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ExecuteRequestUtils {
```
yinan-symphony, line 12: Or simply lomok annotation @UtilityClass

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/request/ExecuteRequestUtils.java
```diff
@@ -0,0 +1,42 @@
...
+        .forEach(key -> {
+          List<String> encodedValues =
+              queryParamsMap.get(key)
+                  .stream()
+                  .map(ExecuteRequestUtils::encode)
+                  .collect(Collectors.toList());
+          clone.replaceQueryParam(key, encodedValues);
+        });
+
+    return clone.build().toUriString();
```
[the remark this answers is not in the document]
yinan-symphony: oh at the end, we have to replace the query param string for sure. UriBuilder.java from java could help

symphony-youri (approved):

PR #144, opened 2022-09-14
Access control for monitoring apis

symphony-soufiane:
### Description
In this PR we added access control to the monitoring apis.
The monitoring apis require now a token "X-Monitoring-Token" in the request header in order to authorise the request.
The token value to match is provided as an environment variable to WDK jar and it is stored as a property in application.yaml with a placeholder. 

In terms of implementation, an annotation Authorized that contains the token key is added. This annotation is handled as a pointcut with AspectJ in order to match tokens.  

For the same occasion, WDK config has been changed: the workflows folder property becomes under wdk.workflows.path and a new property wdk.properties.montoring-token is added for the monitoring api token.

When no token is configured, all monitoring requests will be, by design, unauthorised.

### Dependencies
org.aspectj:aspectjrt:1.9.9.1
org.aspectj:aspectjweaver:1.9.9.1


workflow-bot-app/src/main/java/com/symphony/bdk/workflow/exception/ApiDisabledException.java
```diff
@@ -0,0 +1,9 @@
+package com.symphony.bdk.workflow.exception;
+
+
+public class ApiDisabledException extends RuntimeException {
```
yinan-symphony, line 4: i think this is not necessary anymore, isn't it ?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/exception/GlobalExceptionHandler.java
```diff
@@ -23,6 +23,11 @@ public ResponseEntity<ErrorResponse> handle(Throwable exception) {
     return handle(exception.getMessage(), HttpStatus.INTERNAL_SERVER_ERROR);
   }
 
+  @ExceptionHandler(ApiDisabledException.class)
+  public ResponseEntity<ErrorResponse> handle(ApiDisabledException exception) {
```
yinan-symphony, line 27: same here

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/security/AuthorizationAspect.java
```diff
@@ -0,0 +1,47 @@
...
+  public AuthorizationAspect(@Value("${wdk.properties.monitoring-token}") String monitoringToken) {
+    this.monitoringToken = monitoringToken;
+  }
+
+  @Before("@within(org.springframework.web.bind.annotation.RequestMapping) && @annotation(authorized)")
+  public void authorizationCheck(Authorized authorized) {
+    HttpServletRequest httpServletRequest = getHttpServletRequest();
+
+    if (monitoringToken == null || monitoringToken.isEmpty()) {
+      throw new ApiDisabledException(
```
yinan-symphony, line 32: it suggest to use the UnauthoriedException here, too.

workflow-bot-app/src/test/resources/application-test.yaml
```diff
@@ -1,3 +1,7 @@
+wdk:
```
yinan-symphony, line 1: do we need a configuration class to hold these wdk configs ? i didn't an existing one.
the workflows section should be moved inside as well.

yinan-symphony (approved):
LGTM

PR #147, opened 2022-09-20
WIP: first take on query params handling

symphony-soufiane:
WIP: first take on query params handling

### DIFF
--- a/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/api/v1/dto/StatusEnum.java
+++ b/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/api/v1/dto/StatusEnum.java
@@ -5,6 +5,12 @@ public enum StatusEnum {
   COMPLETED;
 
   public static StatusEnum toInstanceStatusEnum(String status) {
-    return "ACTIVE".equals(status) ? StatusEnum.PENDING : StatusEnum.COMPLETED;
+    if ("ACTIVE".equalsIgnoreCase(status) || PENDING.name().equalsIgnoreCase(status)) {
+      return StatusEnum.PENDING;
+    } else if (COMPLETED.name().equalsIgnoreCase(status)) {
+      return StatusEnum.COMPLETED;
+    } else {
+      throw new IllegalArgumentException(String.format("Workflow instance status %s is not known", status));
+    }
   }
 }
--- a/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/monitoring/repository/ActivityCmdaApiQueryRepository.java
+++ b/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/monitoring/repository/ActivityCmdaApiQueryRepository.java
@@ -1,5 +1,6 @@
 package com.symphony.bdk.workflow.engine.camunda.monitoring.repository;
 
+import com.symphony.bdk.workflow.api.v1.dto.WorkflowInstLifeCycleFilter;
 import com.symphony.bdk.workflow.converter.ObjectConverter;
 import com.symphony.bdk.workflow.monitoring.repository.ActivityQueryRepository;
 import com.symphony.bdk.workflow.monitoring.repository.domain.ActivityInstanceDomain;
@@ -8,10 +9,12 @@ import com.symphony.bdk.workflow.monitoring.repository.domain.VariablesDomain;
 import org.camunda.bpm.engine.HistoryService;
 import org.camunda.bpm.engine.RepositoryService;
 import org.camunda.bpm.engine.RuntimeService;
+import org.camunda.bpm.engine.history.HistoricActivityInstanceQuery;
 import org.camunda.bpm.engine.history.HistoricVariableInstance;
 import org.springframework.stereotype.Component;
 
 import java.util.Collections;
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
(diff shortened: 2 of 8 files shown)


workflow-bot-app/src/main/java/com/symphony/bdk/workflow/api/v1/dto/StatusEnum.java
```diff
@@ -2,9 +2,22 @@
 
 public enum StatusEnum {
   PENDING,
-  COMPLETED;
+  COMPLETED,
+  FAILED;
 
   public static StatusEnum toInstanceStatusEnum(String status) {
-    return "ACTIVE".equals(status) ? StatusEnum.PENDING : StatusEnum.COMPLETED;
+    if (status == null) {
```
yinan-symphony, line 9: i think we can first make the input value all capital, and use StatusEnum.valueOf() to get the instance. it will be much simpler, and the exception is handled also inside the valueOf method

workflow-bot-app/build.gradle
```diff
@@ -59,6 +59,9 @@ dependencies {
     implementation 'io.springfox:springfox-swagger2:2.9.2'
     implementation 'io.springfox:springfox-swagger-ui:2.9.2'
 
+    implementation 'joda-time:joda-time:2.11.1'
```
yinan-symphony, line 62: Maybe using directly java time is better IMO.

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/monitoring/repository/ActivityCmdaApiQueryRepository.java
```diff
@@ -56,7 +83,12 @@ public List<ActivityInstanceDomain> findAllByWorkflowInstanceId(String instanceI
 
     result.stream()
         .filter(a -> a.getType().equals("serviceTask"))
-        .forEach(activity -> activity.setVariables(variablesDomainMap.get(activity.getName())));
+        .forEach(activity -> {
+          VariablesDomain variablesDomain = variablesDomainMap.get(activity.getName());
+          if (variablesDomain != null) {
```
yinan-symphony, line 88: can be done in a filter

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/monitoring/repository/WorkflowInstQueryRepository.java
```diff
@@ -1,9 +1,10 @@
 package com.symphony.bdk.workflow.monitoring.repository;
 
+import com.symphony.bdk.workflow.api.v1.dto.StatusEnum;
 import com.symphony.bdk.workflow.monitoring.repository.domain.WorkflowInstanceDomain;
 
 import java.util.List;
 
 public interface WorkflowInstQueryRepository extends QueryRepository<WorkflowInstanceDomain, String> {
-  List<WorkflowInstanceDomain> findAllById(String id);
+  List<WorkflowInstanceDomain> findAllById(String id, StatusEnum status);
```
yinan-symphony, line 9: maybe we can have two methods, one is findAllById, the other is findAllByIdAndStatus, it would be clearer and avoid passing null value for status arg

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/monitoring/repository/ActivityCmdaApiQueryRepository.java
```diff
@@ -31,11 +35,34 @@ public ActivityCmdaApiQueryRepository(RepositoryService repositoryService,
...
+    if (StringUtils.isNotBlank(lifeCycleFilter.getFinishedAfter())) {
+      historicActivityInstanceQuery.finishedAfter(new DateTime(lifeCycleFilter.getFinishedAfter()).toDate());
+    }
+
+    List<ActivityInstanceDomain> result = objectConverter.convertCollection(historicActivityInstanceQuery
+        .orderByHistoricActivityInstanceStartTime()
+        .asc()
+        .list()
+        .stream()
+        .filter(actInstance -> workflowId.equals(actInstance.getProcessDefinitionKey()))
```
yinan-symphony, line 64: the check is already done at service level, i think this filter is not necessary, right?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/security/AuthorizationAspect.java
```diff
@@ -29,6 +30,10 @@ public void authorizationCheck(Authorized authorized) {
     String headerKey = authorized.headerTokenKey();
     HttpServletRequest httpServletRequest = getHttpServletRequest();
 
+    if (headerKey != null && httpServletRequest.getHeader(headerKey) == null) {
+      throw new UnauthorizedException(String.format(UNAUTHORIZED_EXCEPTION_MISSING_TOKEN_MESSAGE, headerKey));
```
yinan-symphony, line 34: I would suggest to make the error message more generic for security check failing. It is a good practice, like we 'd never explicitly say in a name/pwd authentication that the username is wrong or password is wrong, but simply say credentials wrong.

symphony-soufiane: You're right, I'm now using the same error "Request token is not valid"

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/api/v1/dto/StatusEnum.java
symphony-soufiane: The issue here is that we need to map both "ACTIVE" and "PENDING" to StatusEnum.PENDING. In Camunda's point of view, a process can be "active" while in WDK POV it is "PENDING".  
I also prefer to mention the valid values in the error which is not provided in the exception returned by valueOf(). WDYT?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/security/AuthorizationAspect.java
```diff
@@ -31,7 +30,7 @@ public void authorizationCheck(Authorized authorized) {
     HttpServletRequest httpServletRequest = getHttpServletRequest();
 
     if (headerKey != null && httpServletRequest.getHeader(headerKey) == null) {
-      throw new UnauthorizedException(String.format(UNAUTHORIZED_EXCEPTION_MISSING_TOKEN_MESSAGE, headerKey));
+      throw new UnauthorizedException(UNAUTHORIZED_EXCEPTION_INVALID_TOKEN_MESSAGE);
```
yinan-symphony, line 33: these lines can be removed.

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/security/AuthorizationAspect.java
```diff
@@ -16,7 +16,6 @@
 public class AuthorizationAspect {
 
   private static final String UNAUTHORIZED_EXCEPTION_INVALID_TOKEN_MESSAGE = "Request token is not valid";
```
yinan-symphony, line 18: We can then make the error message more generic, like "Request is not authorised."

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/monitoring/service/MonitoringService.java
```diff
@@ -131,12 +131,12 @@ public WorkflowDefinitionView getWorkflowDefinition(String workflowId) {
...
 
-  public List<VariableView> listWorkflowInstanceGlobalVars(String workflowId, String instanceId, String occurredBefore,
-      String occurredAfter) {
+  public List<VariableView> listWorkflowInstanceGlobalVars(String workflowId, String instanceId, String updatedBefore,
+      String updatedAfter) {
     // check if the instance belongs to the provided workflow
     this.checkIsInstanceOfWorkflow(workflowId, instanceId);
 
-    return variableQueryRepository.findGlobalVarsHistoryByWorkflowInstId(instanceId, occurredBefore, occurredAfter)
+    return variableQueryRepository.findGlobalVarsHistoryByWorkflowInstId(instanceId, updatedBefore, updatedAfter)
```
yinan-symphony, line 139: should be better to parse the string into Instant at controller level, so that for the rest of code, we use instant, only in repository we parse it to Date. I think we can directly use Instant type for the query params in the controller

yinan-symphony (approved):

Issue #149, opened 2022-09-27
WDK 1.2.0: Send Message action not initiating for 'One-Of' activity when 'Request-Received' + 'Form-Replied'


Kaosar-Symphony:
**WDK 1.2.0: Send Message action not initiating for 'One-Of' activity when 'Request-Received + 'Form-Replied'**

### Steps to Reproduce:
1. Open WDK 1.2.0 + the SWADL (https://github.com/SymphonyPlatformSolutions/symphony-training-research-workflow/blob/chapter-6.1/workflows/research.swadl.yaml)

2. Use Postman to initiate the webhook URL (http://localhost:8080/wdk/v1/workflows/research-webhook-workflow/execute)
Header = X-Workflow-Token
Body = { "args": { "webhookTicker": "AAPL" , "webhookContent": "Apple Inc. has released new popular mobile devices and laptops. There is speculation that Apple will release a self-driving car by 2025. We should consider this stock for the future." }}

3. In the Symphony Pod Environment - check the conversation chatroom as per the SWADL. 

### Expected Result:
The webhook from Postman should initiate the WDK bot to create a conversation with the end-user and send an element form in the Symphony Pod Environment.

### Actual Result:
In WDK 1.2.0 - This does not work - the Send Message activity in the WDK Bot logs show that it has received the webhook but has not sent the message.
In WDK 1.1.0 - This works - The Send Message activity is mentioned in the WDK Bot logs along with the 'request-received' webhook - with the elements form showing up in the Symphony Pod Environment.

### Environment:
**WDK 1.2.0**
**SWADL** = (https://github.com/SymphonyPlatformSolutions/symphony-training-research-workflow/blob/chapter-6.1/workflows/research.swadl.yaml)

### Additional Context:
It is important to note that tweaking the SWADL to remove the 'One-Of' activity and keeping just the single event trigger 'request-received' - allows for the message to be sent. However, there have been instances where the 'One-of' and the 'Form-Replied' might be conflicting which is why the Send Message does not action.

symphony-soufiane:
Fixed and will be released in WDK 1.3.0
<img width="817" alt="149" src="https://user-images.githubusercontent.com/52406574/195353896-3ac3eef6-2b56-4932-adba-4f8274b2f058.png">

PR #151, opened 2022-09-27
#149: Fix bug when on request-received event is added with a subsequent activity having form-replied event.

symphony-soufiane:
### Description
#149 
The issue happens when the workflow has an activity with triggering event request-received within one-of property that sends a form, and a subsequent activity executed on that form replied. The workflow being built has this request-received event attached to an endEvent which makes the workflow completes once the request is received without executing the activities. The issue comes from an incomplete condition checking if the activity is in a loop. We enriched this condition by checking if the current activity is of type ACTIVITY_COMPLETED_EVENT.

**Before the fix:**
Screenshot 2022-09-26 at 20 12 24

**After the fix:**
Screenshot 2022-09-27 at 17 28 12

### DIFF
--- a/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/bpmn/builder/FormRepliedNodeBuilder.java
+++ b/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/bpmn/builder/FormRepliedNodeBuilder.java
@@ -39,7 +39,7 @@ public class FormRepliedNodeBuilder extends AbstractNodeBpmnBuilder {
       SubProcessBuilder subProcess = builder.subProcess();
       context.cacheSubProcess(subProcess);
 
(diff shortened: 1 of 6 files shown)


workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/bpmn/builder/FormRepliedNodeBuilder.java
```diff
@@ -39,7 +39,7 @@ public class FormRepliedNodeBuilder extends AbstractNodeBpmnBuilder {
       SubProcessBuilder subProcess = builder.subProcess();
       context.cacheSubProcess(subProcess);
 
-      timeoutFlow(element, subProcess, context);
```
symphony-soufiane, line 42: not related to the issue: unused parameter

yinan-symphony (approved):

PR #155, opened 2022-09-29
Fix bug form replied and if condition

symphony-soufiane:
### Description

### Dependencies

### Checklist


workflow-bot-app/src/test/java/com/symphony/bdk/workflow/BranchingIntegrationTest.java
```diff
@@ -67,6 +75,24 @@ void branching(String workflowFile, List<String> activities) throws IOException,
     assertThat(workflow).isExecutedWithProcessAndActivities(lastProcess(), activities);
   }
 
+  @Test
+  void branching_intermediate() throws IOException, ProcessingException, InterruptedException {
```
symphony-soufiane, line 79: isn't this test similar to the one done here? 
test: https://github.com/finos/symphony-wdk/blob/822cacf4b162055880dd89fe1c1c8afb98b728e6/workflow-bot-app/src/test/java/com/symphony/bdk/workflow/BranchingIntegrationTest.java#L37
swadl: https://github.com/finos/symphony-wdk/blob/822cacf4b162055880dd89fe1c1c8afb98b728e6/workflow-bot-app/src/test/resources/branching/if-else-if.swadl.yaml



We would prefer to test with an unsatisfied condition (if: variables.foo != "foo") rather than (.. == "foo") to verify the condition works. (or both cases actually)

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/FormReplyIntegrationTest.java
```diff
@@ -323,4 +322,53 @@ void sendMessageUpdateMessage() throws Exception {
...
+
+    // trigger workflow execution
+    engine.onEvent(messageReceived("/test"));
+    verify(messageService, timeout(5000)).send(anyString(), contains("form"));
+
+    await().atMost(5, TimeUnit.SECONDS).ignoreExceptions().until(() -> {
+      engine.onEvent(form("msgId", "testForm", Collections.singletonMap("action", "create")));
+      return true;
+    });
+    verify(messageService, timeout(5000)).send(anyString(), contains("Create"));
```
symphony-soufiane, line 343: I'm not sure this verifies what you think. 
In line 337, we verify that we sent the form (content contains the word form). 
In line 343, we want to verify that we sent a the message "Create". I don't know why Mockito.verify does not get updated, and it gets applied on the first one. So both Mockito.verify applies on the first messageService.send(..), which contains both "form" and "Create" words. 

If you change the activity "resCreate" in swadl to send something else than "Create", let's say "CreateABC", and you change the test with (contains("CreateABC")), the test will fail.. One way to force Mockito.verify to apply to the new sendMessage call, is to change the to.streamId="" to a different one and replace anyString() in verify(..). 


```verify(messageService, timeout(5000)).send("ANOTHER_STREAM_ID", contains("SOMETHING_DIFF_FROM_CREATE"));```

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/BranchingIntegrationTest.java
yinan-symphony: 👍  new test is removed, it is indeed kinda duplicated, so keep the existing test only.

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/FormReplyIntegrationTest.java
yinan-symphony: ok, i see your point, i add then the Mockito.clearInvocations(T mock) to clear the invocations on messageService mock after each verify, so that we are sure we have the right verification.

yinan-symphony (approved):

PR #170, opened 2022-10-14
Improve the workflow definition structure

yinan-symphony:
This improvement contains
 - add the gateways in the definition
 - add condition string at child level if there is any

### Description

### Dependencies

### Checklist


workflow-bot-app/src/main/java/com/symphony/bdk/workflow/monitoring/service/MonitoringService.java
```diff
@@ -100,42 +112,95 @@ public WorkflowActivitiesView listWorkflowInstanceActivities(String workflowId,
...
+  @Nullable
+  private static String determineCondition(Map<String, WorkflowNode> dictionary, String key, String node) {
+    WorkflowNode child = dictionary.get(node);
+    String condition = child.getIfCondition(key);
+    if (condition == null) {
+      condition = child.getElementType() == WorkflowNodeType.ACTIVITY_EXPIRED_EVENT ? "expired"
+          : child.getElementType() == WorkflowNodeType.ACTIVITY_FAILED_EVENT ? "failed" : null;
+    }
+    return condition;
+  }
```
symphony-soufiane, line 202: I suggest to directly pass the child instead of dictionary+node, so we only pass the required objects.
```suggestion
@Nullable
private static String determineCondition(WorkflowNode child, String key) {
    String condition = child.getIfCondition(key);
    if (condition == null) {
      condition = child.getElementType() == WorkflowNodeType.ACTIVITY_EXPIRED_EVENT ? "expired"
          : child.getElementType() == WorkflowNodeType.ACTIVITY_FAILED_EVENT ? "failed" : null;
    }
    return condition;
  }
```

workflow-bot-app/src/test/resources/monitoring/oneof-allof-gateway-definition.swadl.yaml
```diff
@@ -0,0 +1,36 @@
...
+        message-received:
+          content: /start
+          # should be 1-1 only
+      content: start fork
+
+  - execute-script:
+      id: scriptTrue
+      if: ${variables.allOf == true}
+      script:
+        println("script allOf true is done")
```
symphony-soufiane, line 17: no need to print a message, we can just put an empty script here

workflow-bot-app/src/test/resources/monitoring/oneof-allof-gateway-definition.swadl.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+  - execute-script:
+      id: scriptFalse
+      on:
+        activity-completed:
+          activity-id: start
+      script:
+        println("script allOf false is done")
```
symphony-soufiane, line 25: same

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/monitoring/service/MonitoringService.java
yinan-symphony: > I suggest to directly pass the child instead of dictionary+node, so we only pass the required objects.

sure. agree, looks like Intellji extract code is not that intelligent

yinan-symphony: @symphony-soufiane done as suggested

Issue #156, opened 2022-09-29
Enhance blast messaging processing


richardring:
### Description of Problem:
When sending a message to multiple rooms there isn't a way to interrogate the rooms you sent afterwards in downstream workflows, you can only access the first message ID. 

### Potential Solutions:
Access as a list the message IDs from the activity?

PR #174, opened 2022-10-19
#156: Enhance blast message and support multiple events

symphony-soufiane:
### Description
#156 Enhance blast message and support multiple events. 
When sending a blast message, some messages might fail. The api response contains the failing ones in an error object and the successful ones. So far, WDK was only storing the first successful message in send-message activity  outputs. 
In this PR, we add new outputs for the send-message activity to store the list of successful messages and the list of failing ones. The existing outputs are kept along with the new ones.
We also changed the behaviour of send-message activity when a list of user ids is provided. So far, an MIM with those uses was created or reused. We now send the message in an IM to each user, in a blast fashion. 
The particularity of send-message with user-ids list remains in the fact that we convert each user-id to a stream id (the corresponding IM). In case of an invalid user-id, the api returns 403 error. In order, to keep processing the rest of user-ids, we catch this error and ignore it.

### Checklist
- [x] Javadoc added or updated
- [x] Updated the documentation in [docs folder](../docs)


workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/message/SendMessageExecutor.java
```diff
@@ -42,6 +43,8 @@ public class SendMessageExecutor extends OboExecutor<SendMessage, V4Message>
   // required for message correlation and forms (correlation happens on variables than cannot be nested)
   public static final String OUTPUT_MESSAGE_ID_KEY = "msgId";
   public static final String OUTPUT_MESSAGE_KEY = "message";
+  public static final String OUTPUT_MESSAGES_KEY = "messages";
+  public static final String OUTPUT_FAILED_MESSAGES_KEY = "failedStreamIds";
```
symphony-soufiane, line 47: We might want to improve this field name. Any idea @yinan-symphony ?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/executor/message/SendMessageExecutor.java
```diff
@@ -122,23 +139,30 @@ private List<String> resolveStreamId(ActivityExecutorContext<SendMessage> execut
...
-    return stream.getId();
+  private List<String> createOrGetStreamId(List<Long> userIds, StreamService streamService) {
+    List<String> streamIds = new ArrayList<>();
+
+    for (Long userId : userIds) {
+      // passing a singleton list of long instead of long to make the test mocking easy
+      try {
+        streamIds.add(streamService.create(List.of(userId)).getId());
+      } catch (ApiRuntimeException apiRuntimeException) {
+        // ignore error when user is not found
```
symphony-soufiane, line 158: @yinan-symphony WDYT? I am not really happy of this handling, do you have a better suggestion?

yinan-symphony: do you mean the ignore on the error when the user is not found? 
yes, it would be great if we can make it traceable, like put the user id in a variable. 
I think we are missing a generic error variable, which holds the potential execution exceptions.
btw, please add a error log for the exception message only, and a debug log for the exception stacktrace.

PR #180, opened 2022-10-27
#156 Enhance blast message and support multiple events 

### LINKED ISSUE #156: Enhance blast messaging processing

symphony-soufiane:
**Enhance blast message**
So far, when sending a blast message, only the first messageId was stored in activity outputs. In this PR, we introduced two new outputs storing the list of message Ids, and the list of the failing ones. 
For the same occasion, when a send-message activity has a list of user-ids, a blast message is sent to each user instead of creating a room (MIM) with all of them. User ids are converted to stream ids: a private message with the bot (IM) if exists, otherwise a new stream is created. 

**Support multiple events**
So far, only the latest event being received was stored in variables and could be accessed with the key _"event"_. This is due to the fact that, every time an event is received, a variable with the key "event" was created, therefore the previous event is overridden. In this PR, we changed this behaviour to store events with a unique key in order to keep all of them. For that, an optional id in introduced in SWADL for events, when it is set and the event is received, a new variable is stored. The event can be accessed in SWADL using this id. Please note that the "id" property is not unique and two events can use the same one, in this case the last one to be received will be stored. Fix this issue in a clean way would required a refactoring WDK code planned in a next PR.

closes #156


workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/WorkflowNode.java
```diff
@@ -20,6 +20,7 @@
 public class WorkflowNode {
   private String id;
   private Event event;
+  private String eventId = "tmp";
```
yinan-symphony, line 23: maybe the default value here is not necessary, in any case, during the graph construction, the value is going to be replaced by the real event id or the same value as id in here.

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/WorkflowEventToCamundaEvent.java
```diff
@@ -178,72 +180,115 @@ public <T> void dispatch(RealTimeEvent<T> event) throws PresentationMLParserExce
 
     } else if (event.getSource() instanceof V4RoomCreated) {
       log.debug("receive a room created event");
+
+      ((EventHolder) processVariables.get(ActivityExecutorContext.EVENT)).getArgs()
+          .put(EVENT_NAME, WorkflowEventToCamundaEvent.ROOM_CREATED);
```
yinan-symphony, line 185: create a function and pass just the constant, we may save a bit long line code

yinan-symphony: even for the next two lines, they can extract to a method together

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/WorkflowEventHandler.java
```diff
@@ -0,0 +1,86 @@
...
+
+        if (StringUtils.isNotBlank(eventId)) {
+          // remove event name from args
+          eventHolder.getArgs().remove(WorkflowEventToCamundaEvent.EVENT_NAME);
+
+          // store the event with the new key
+          this.runtimeService.setVariable(event.getExecutionId(), eventId, eventHolder);
+        }
+      } catch (JsonProcessingException e) {
+        e.printStackTrace();
```
yinan-symphony, line 81: not stacktrace

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/WorkflowEventHandler.java
```diff
@@ -0,0 +1,86 @@
...
+import org.camunda.bpm.engine.impl.history.event.HistoryEvent;
+import org.camunda.bpm.engine.impl.history.handler.HistoryEventHandler;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.stereotype.Component;
+
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+
+@Component
+public class WorkflowEventHandler implements HistoryEventHandler {
```
yinan-symphony, line 23: i kinda like to make this class as an event publisher, using spring publisher, and let AuditTrailLogger and another class to deal respectively with audit trail log and variables, it is more extensible if one day we have other need on this history event, in addition, it will get rid of the cycle dependency, as they are completely decoupled.

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/WorkflowNode.java
symphony-soufiane: Yes I should not have committed this one, it was only for my testing

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/WorkflowEventHandler.java
symphony-soufiane: I do prefer strategy pattern actually, I will give it a try in next commit.

yinan-symphony: 👍 , now it s much better.

yinan-symphony (approved):

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/WorkflowEventToCamundaEvent.java
```diff
@@ -304,7 +304,14 @@ private <T> void formReplyToMessage(RealTimeEvent<T> event, Map<String, Object>
   @SuppressWarnings({"unchecked", "rawtypes"})
   private void requestReceivedToRequest(RequestReceivedEvent eventSource, Map<String, Object> processVariables) {
     String eventName = requestReceivedWorkflowEvent(eventSource.getWorkflowId());
-    Map<String, Object> args = new HashMap<>(eventSource.getArguments());
+    Map<String, Object> args;
+
+    if (eventSource.getArguments() == null) {
```
yinan-symphony, line 309: Suggest it to be like

`Map<String, Object> args = new HashMap<>();

    if (eventSource.getArguments() != null) {
      args.putAll(eventSource.getArguments());
    }`

yinan-symphony (approved):

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/engine/camunda/WorkflowEventToCamundaEvent.java
symphony-soufiane: eventSource.getArguments() is immutable, that's why I'm cloning it in the "else" .

PR #181, opened 2022-10-27
Add events in the payload of the workflow endpoint

yinan-symphony:
- Change the workflow URL from /activites to /states to present the current state of the workflow instance

 - Add events in the changed endpoint, while only activities were added in the payload before.

### Description

### Dependencies

### Checklist

### DIFF
--- a/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/WorkflowValidator.java
+++ b/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/WorkflowValidator.java
@@ -15,10 +15,6 @@ import org.apache.commons.lang3.StringUtils;
 public class WorkflowValidator {
 
   public static void validateFirstActivity(BaseActivity activity, Event event, String workflowId) {
-    if (activity.getIfCondition() != null) {
-      throw new InvalidActivityException(workflowId,
-          String.format("Workflow's starting activity %s must not have a conditional branching", activity.getId()));
-    }
     if (event instanceof EventWithTimeout && StringUtils.isNotEmpty(((EventWithTimeout) event).getTimeout())) {
       throw new InvalidActivityException(workflowId,
           String.format("Workflow's starting activity %s must not have timeout", activity.getId()));
--- a/workflow-bot-app/src/test/java/com/symphony/bdk/workflow/BranchingIntegrationTest.java
+++ b/workflow-bot-app/src/test/java/com/symphony/bdk/workflow/BranchingIntegrationTest.java
@@ -40,13 +40,8 @@ class BranchingIntegrationTest extends IntegrationTest {
   @SuppressWarnings("checkstyle:LineLength")
   static Stream<Arguments> expectedErrors() {
     return Stream.of(
-        arguments("/branching/invalid/swadl/if-starting-activity.swadl.yaml", InvalidActivityException.class,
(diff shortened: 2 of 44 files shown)


workflow-bot-app/src/main/java/com/symphony/bdk/workflow/WorkflowValidator.java
```diff
@@ -15,10 +15,6 @@
 public class WorkflowValidator {
 
   public static void validateFirstActivity(BaseActivity activity, Event event, String workflowId) {
-    if (activity.getIfCondition() != null) {
```
symphony-soufiane, line 18: I just noticed we don't prohibit "allOf" in this check. Actually when the first activity has a "allOf", the next activities will never be executed. (I was expecting new instance to start but the first one to have allOf events fulfilled and move forward, but it's worst).

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/BranchingIntegrationTest.java
```diff
@@ -40,8 +40,8 @@ static Stream<Arguments> executedActivities() {
   @SuppressWarnings("checkstyle:LineLength")
   static Stream<Arguments> expectedErrors() {
     return Stream.of(
-        arguments("/branching/invalid/swadl/if-starting-activity.swadl.yaml", InvalidActivityException.class,
-            "Invalid activity in the workflow if-in-starting-activity: Workflow's starting activity startingActivity must not have a conditional branching"),
+        //arguments("/branching/invalid/swadl/if-starting-activity.swadl.yaml", InvalidActivityException.class,
```
symphony-soufiane, line 43: It's better to remove this test, or change it to not fail when starting activity has if condition

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/WorkflowValidatorTest.java
```diff
@@ -47,14 +43,6 @@ static Stream<Arguments> firstActivities() {
     );
   }
 
-  @ParameterizedTest
-  @MethodSource("firstActivities")
```
symphony-soufiane, line 51: Either delete the stream firstActivities() as it is not used anymore,  or keep it and make the test to pass instead of failing, to cover this case where starting activity has a condition

symphony-soufiane (approved):

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/BranchingIntegrationTest.java
yinan-symphony: removed

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/WorkflowValidatorTest.java
yinan-symphony: removed

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/WorkflowValidator.java
yinan-symphony: i will test this case to validate the behavior.

PR #185, opened 2022-11-02
Support timeout in monitoring api instances endpoints

symphony-soufiane:
**Support timeout in monitoring api**
In this PR we added the support of timeout events in the workflow instance monitoring api. It was missing because the wrappedType attribute was not set in the WorkflowNode of a timeout event. This was causing a NP during the monitoring processing. 

**BDK upgrade**
For the same occasion, the BDK version has been upgraded to the latest 2.10.0. This version of BDK comes with the upgrade of spring-boot-dependencies where H2 database version was upgraded. This version of H2 is not compatible with Camunda, so the version has been strictly put in the build gradle to keep using h2 database 1.4.200.


ystxn (approved):
LGTM

PR #188, opened 2022-11-10
Add workflow SWADL management API

yinan-symphony:
WDK management API allows to deploy, update, delete workflow SWADL files, instead of dropping SWADL files to the disk manually.

### Description

### Dependencies

### Checklist


workflow-bot-app/src/main/java/com/symphony/bdk/workflow/management/WorkflowAbstractAction.java
```diff
@@ -0,0 +1,61 @@
...
+  protected Workflow convertToWorkflow(String content) {
+    try {
+      return SwadlParser.fromYaml(content);
+    } catch (Exception e) {
+      throw new IllegalArgumentException("SWADL content is not valid");
+    }
+  }
+
+  protected void validateFilePath(String id, String path) {
+    if (deployer.workflowSwadlPath(id) != null && deployer.workflowSwadlPath(id).endsWith(path)) {
```
symphony-soufiane, line 32: isn't `deployer.workflowSwadlPath(id) != null` enough? if a workflow is already deployed with "id",  we should throw DuplicateException.

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/management/WorkflowAbstractAction.java
```diff
@@ -0,0 +1,61 @@
...
+    return deployer.workflowExist(id);
+  }
+
+  protected void writeFile(String content, Workflow workflow, String path) {
+    try {
+      BufferedWriter writer = new BufferedWriter(new FileWriter(path, StandardCharsets.UTF_8));
+      writer.write(content);
+      writer.close();
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to write SWADL file " + workflow.getId());
```
symphony-soufiane, line 51: We can maybe tell more about the exception in the error message?

symphony-soufiane (commented):
Can you update the postman collection please?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/management/WorkflowAbstractAction.java
yinan-symphony: it s not enough, this check is to avoid overwriting a file having the same file name but different workflow id (e.g. manually dropped in the workflow directory), and the file's name happens to be the same with the received workflow id. The case would be very rare, but could happen.

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/management/WorkflowAbstractAction.java
yinan-symphony: 👍  sure

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/management/WorkflowAbstractAction.java
yinan-symphony: i changed this check, the way was done before was indeed not correct.

PR #192, opened 2022-11-24
Fix custom activity type unknown issue in definitino endpoint

yinan-symphony:
The activity type and group were predefined in an enum class, that prevent the definition endpoint from knowing a custom activity's type and group (with null value).

This commit fix the issue. Instead of doing the mapping through the enum class, now it is dynamically reading and converting from the activity class simple name.

### Description

### Dependencies

### Checklist


workflow-bot-app/build.gradle
```diff
@@ -62,6 +62,7 @@ dependencies {
 
     implementation 'com.github.ben-manes.caffeine:caffeine:3.1.1'
     implementation 'io.springfox:springfox-boot-starter:3.0.0'
+    implementation 'com.google.guava:guava:31.1-jre'
```
symphony-soufiane, line 65: Does it worth adding a whole new library only for formatting the activity type name? Can't we do it with a utility function without this dependency ?

workflow-language/src/main/java/com/symphony/bdk/workflow/swadl/v1/Event.java
```diff
@@ -13,7 +13,6 @@
 import com.symphony.bdk.workflow.swadl.v1.event.RequestReceivedEvent;
 import com.symphony.bdk.workflow.swadl.v1.event.RoomCreatedEvent;
 import com.symphony.bdk.workflow.swadl.v1.event.RoomDeactivatedEvent;
-import com.symphony.bdk.workflow.swadl.v1.event.RoomMemberDemotedFromOwnerEvent;
```
symphony-soufiane, line 16: Can you please change the field type in ligne 79 to RoomMemberDemotedFromOwnerEvent?

workflow-bot-app/src/test/java/com/symphony/bdk/workflow/CustomActivityIntegrationTest.java
```diff
@@ -19,6 +29,18 @@ void customActivity() throws Exception {
 
     // com.symphony.bdk.workflow.DoSomethingExecutor is just sending a message
     verify(messageService, timeout(5000)).send("123", "abc");
+
+    Response response = given()
```
symphony-soufiane, line 33: I'm wondering if we should add tests here or add custom activity based tests in the monitoring api tests class.

yinan-symphony: It might not make a big difference, it is a matter of organisation. IMO, the custom activity is quite specific, I prefer to put related stuff in one place.

workflow-language/src/main/java/com/symphony/bdk/workflow/swadl/v1/Event.java
yinan-symphony: sure.

workflow-bot-app/build.gradle
yinan-symphony: i will have a look.

yinan-symphony: done

symphony-soufiane (approved):

PR #203, opened 2022-12-12
Standard debug Activity

symphony-soufiane:
### Description
Introduced Debug activity to be used to inspect object values in a SWADL. 
The debug activity does not require an Id in which case a random one is set.

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -91,6 +91,27 @@ dozen lines maybe a custom activity should be considered instead.
 
 You'll find various usages of it in the [examples](./examples) we provide.
 
+### Debug Activity
+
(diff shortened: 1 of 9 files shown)


ystxn (approved):
LGTM

yinan-symphony:
@symphony-soufiane should we remove this Debug activity from the nodes list endpoint? as it is for debug purpose only.

yinan-symphony (approved):

ystxn:
> @symphony-soufiane @ystxn should we remove this Debug activity from the nodes endpoint response? as it is for debug purpose only.

no need to specially exclude. since this sets the output variable, it could be used in the monitoring endpoints to inspect an event etc, so it should be consistent with the nodes list.

PR #205, opened 2022-12-16
#198 Expose Bot User Profile

### LINKED ISSUE #198: Expose Bot User Profile
### Description of Problem:
There is currently no easy way to organically obtain the current service account's user profile (user id, display name etc). The only method within SWADL is using `get-users`, but the assumption is another piece of identity like email must be provided, which defeats the purpose. The BDK has a `SessionService` that exposes the bot user profile, so it is possible to create a custom activity like below, though it seems overkill to have to write a custom activity to achieve this.

```java
public class WhoAmI extends BaseActivity {}

public class WhoAmIExecutor implements ActivityExecutor<WhoAmI>  {
    @Autowired
    private final SessionService session;

    @Override
    public void execute(ActivityExecutorContext<WhoAmI> context) {
        context.setOutputVariable("session", session.getSession());
    }
}
```

```yaml
- who-am-i:
    id: whoAmI

- execute-script:
    id: printUserId
    script: println whoAmI.outputs.session.id
```

### Potential Solutions:
Introduce a way in SWADL to obtain the bot user profile, for example:
* A new activity, similar to the above
* A new utility function / object e.g. `wdk.getSession()` or `wdk.session`

symphony-soufiane:
### Description
closes #198 
In this PR, we enriched the Utility functions with a new method allowing to retrieve the bot session info. For this purpose, a static sessionService field is added to the class, the SessionService bean is passed in the public constructor and set to the static field in order to be used in the static method session().

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -164,8 +164,7 @@ defined to express an if/else structure. The default execution in case condition
 too (i.e. an else).
 
 #### Examples
-- [connection admin approval](./examples/connection-admin-approval.swadl.yaml)
-- 
+- [connection admin approval](./examples/connection-admin-approval.swadl.yaml) 
 
 #### Loops
 
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -2529,3 +2529,28 @@ activities:
       script: |
         println wdk.escape(variables.textToEscape)
 ````
+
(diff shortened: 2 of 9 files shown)


ystxn (approved):
LGTM

PR #210, opened 2023-01-19
Support blast form replies

symphony-soufiane:
### Description
In this PR, we added the support of blast form replies.
Having an activity to be executed on a form reply event, if this form has been sent in a blast way, then the activity triggers only when the first form is replied and not the other blast ones.

When a blast message is sent, we only store the first messageId in outputs. This is stored as a variable of the ongoing process instances. 
In the form reply event processing class, we only execute ongoing processes that have a variable msgId equals to the message id of the form that has been actioned. So when the messageId is not the one that is stored in the variable, the process is not executed. 

The fix consists of introducing a new output variable msgIds to the SendMessageExecutor where we store the list of message ids in case of a blast, and the only message id in case it is not a blast. In the form reply event processing class, we fetch for a process id where the variables msgIds list contains the message id of the form that has been actioned and we execute it by its process id. In case no process is found, which can be the case when the on:form-reply event is the starting event of the workflow, then we do not apply this filter which will start a new process execution.

### DIFF
--- a/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/event/V4ElementActionEventProcessor.java
+++ b/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/event/V4ElementActionEventProcessor.java
@@ -9,12 +9,13 @@ import com.symphony.bdk.workflow.engine.executor.message.SendMessageExecutor;
 import lombok.extern.slf4j.Slf4j;
 
 import org.camunda.bpm.engine.RuntimeService;
+import org.camunda.bpm.engine.runtime.MessageCorrelationBuilder;
 import org.camunda.bpm.engine.runtime.VariableInstance;
 import org.springframework.stereotype.Service;
 
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
+import java.util.Optional;
 
 import static java.util.Collections.singletonMap;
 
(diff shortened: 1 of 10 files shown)


workflow-bot-app/src/main/java/com/symphony/bdk/workflow/event/V4ElementActionEventProcessor.java
```diff
@@ -34,10 +37,38 @@ protected void processEventSource(V4SymphonyElementsAction eventSource, Map<Stri
...
+   * @param formId on which the action is applied.
+   * @param messageId of the form.
+   * @return process instance id to be resumed.
+   */
+  private String getProcessToExecute(String formId, String messageId) {
+    List<String> collect = runtimeService.createVariableInstanceQuery().list().stream()
+            .filter(a -> a.getName().equals(String.format("%s.%s.%s", formId, ActivityExecutorContext.OUTPUTS,
+                    SendMessageExecutor.OUTPUT_MESSAGE_IDS_KEY)) && ((List) a.getValue()).contains(messageId))
+            .map(VariableInstance::getProcessInstanceId)
+            .collect(Collectors.toList());
```
yinan-symphony, line 66: could use findFirst to get the single result

yinan-symphony (commented):
it looks like our code formatting is not aligned

yinan-symphony (approved):

PR #213, opened 2023-01-25
Workflow expiration scheduling

symphony-soufiane:
In this PR, we added an api to schedule workflows expiration. 
A thread pool executor is used to enqueue expiration tasks with a given workflow id and the timestamp, when the api is called. When the wdk app restarts, the queue is cleaned as it is memory based, consequently the tasks are lost. To fix that, a new table is added in the wdk database where the required info are stored on an api call and are used to bootstrap the queue when the wdk app starts again.


workflow-bot-app/src/main/java/com/symphony/bdk/workflow/expiration/WorkflowExpirationService.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.stream.Collectors;
+
+@Component
+@ConditionalOnPropertyNotEmpty("wdk.properties.management-token")
+@Transactional(propagation = Propagation.REQUIRES_NEW)
+@Slf4j
+public class WorkflowExpirationService implements WorkflowExpirationInterface {
+  private static final String WORKFLOW_NOT_EXIST_EXCEPTION_MSG = "Workflow %s does not exist.";
+  private final WorkflowExpirationJobRepository expirationJobRepository;
+  protected final VersionedWorkflowRepository versioningRepository;
```
yinan-symphony, line 30: private maybe ?

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/expiration/WorkflowExpirationService.java
```diff
@@ -0,0 +1,71 @@
...
+  private final WorkflowEngine<BpmnModelInstance> workflowEngine;
+
+  private final ScheduledJobsRegistry scheduledJobsRegistry;
+
+  public WorkflowExpirationService(WorkflowExpirationJobRepository workflowExpirationJobRepository,
```
yinan-symphony, line 35: Another option is to use lombok annotation,  very convenient

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/expiration/WorkflowExpirationInterface.java
```diff
@@ -0,0 +1,11 @@
+package com.symphony.bdk.workflow.expiration;
+
+import com.symphony.bdk.workflow.versioning.model.WorkflowExpirationJob;
+
+import java.time.Instant;
+
+public interface WorkflowExpirationInterface {
```
yinan-symphony, line 7: Use Interface as part of the name is not very ideal IMO,  could you change to WorkflowExpirationService, and change the current WorkflowExpirationService implementation to DefaultWorkflowExpirationService.

workflow-bot-app/src/main/java/com/symphony/bdk/workflow/expiration/WorkflowExpirationService.java
```diff
@@ -0,0 +1,71 @@
...
+            workflow.getDeploymentId(), instant))
+        .collect(Collectors.toList());
+
+    if (expirationJobs.isEmpty()) {
+      throw new NotFoundException(String.format(WORKFLOW_NOT_EXIST_EXCEPTION_MSG, workflowId));
+    }
+
+    expirationJobRepository.saveAll(expirationJobs);
+    expirationJobs.forEach(this::scheduleJob);
+  }
```
yinan-symphony, line 58: WorkflowExpirationInterface may not be useful, so leave this method in this service class, then change WorkflowExpirationInterface to something like WorkflowExpirationPlanner, which is an interface, and has one single method, 
`planWorkflowExpiration(WorkflowExpirationJob workflowExpirationjob)`.

yinan-symphony (approved):

PR #224, opened 2023-02-13
Fix response code when trying to update a published workflow

symphony-soufiane:
### Description
The error message returned when trying to update a workflow that has an already published version was confusing at it says that the workflow does not exist. Ths has been fixed in this PR by firstly checking if the workflow exists, then returning a more detailed exception when a version is already published.

### DIFF
--- a/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/management/WorkflowManagementService.java
+++ b/workflow-bot-app/src/main/java/com/symphony/bdk/workflow/management/WorkflowManagementService.java
@@ -27,6 +27,7 @@ import java.util.Optional;
 @Slf4j
 public class WorkflowManagementService {
   private static final String WORKFLOW_NOT_EXIST_EXCEPTION_MSG = "Workflow %s does not exist.";
+  private static final String WORKFLOW_UPDATE_FORBIDDEN_EXCEPTION_MSG = "Update on a published Workflow is forbidden.";
   private final WorkflowEngine<CamundaTranslatedWorkflowContext> workflowEngine;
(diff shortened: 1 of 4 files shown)


workflow-bot-app/src/main/java/com/symphony/bdk/workflow/management/WorkflowManagementService.java
```diff
@@ -101,14 +102,17 @@ public Optional<VersionedWorkflowView> get(String id, Long version) {
...
-    Optional<VersionedWorkflow> activeVersion = versionRepository.findByWorkflowIdAndPublishedFalse(workflow.getId());
+    List<VersionedWorkflow> activeVersion = versionRepository.findByWorkflowId(workflow.getId());
     if (activeVersion.isEmpty()) {
       throw new NotFoundException(String.format(WORKFLOW_NOT_EXIST_EXCEPTION_MSG, workflow.getId()));
     }
-    if (activeVersion.get().getPublished()) {
-      throw new IllegalArgumentException("Update on a published Workflow is forbidden.");
+
+    if (activeVersion.stream().anyMatch(VersionedWorkflow::getPublished)) {
+      throw new UnsupportedOperationException(WORKFLOW_UPDATE_FORBIDDEN_EXCEPTION_MSG);
```
symphony-soufiane, line 111: I'm not sure what is the best Http code to return in this case. 405 does not look suitable as the Method itself is allowed but not on this resource. @ystxn

ystxn: sounds good

ystxn (approved):

Issue #226, opened 2023-02-23
Standard Docker Image


ystxn:
### Description of Problem:
Getting started with WDK development currently requires tooling like JDK installed and also node/npm to get generator to produce a scaffold project. JDK is used to launch the gradle botJar task to download the desired WDK fat jar artifact and launch it. There isn't a real need to touch the project if no custom activities are required.

### Potential Solutions:
As part of the WDK release cycle, a standard Docker image should be published as a GitHub package. This enables developers to run WDK from scratch with a single command without requiring tooling or a java project, assuming they do not wish to develop custom activities.

The standard image should have:
* Tag corresponding to the WDK version
* Documentation for where to mount configuration, keys, libraries
* Examples for disk-based developer mode (workflows directory mount required) or deployment mode (network port proxying required)

Example Dockerfile @ https://github.com/SymphonyPlatformSolutions/wdk-studio/blob/main/wdk-bot/Dockerfile

Additionally, it would be nice to also include the fat jar as part of the release notes so that anyone can download it off a browser rather than have to use the gradle task.

PR #234, opened 2023-03-06
[CICD] - Fixed Docker image push job in the release workflow

symphony-soufiane:
#226 
The docker image push job depends on the release job but it needs to checkout the repository when it starts which makes it lose the fatJar procuded by the previous job. In this commit, we added actions to the release job to download the fatJar as an artifact and download it in the next job to create the docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM openjdk:11-slim-buster AS builder
-COPY workflow-bot-app/build/libs/*.jar wdk.jar
+COPY artifact/*.jar wdk.jar
 
 # Create custom JRE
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM openjdk:11-slim-buster AS builder
-COPY workflow-bot-app/build/libs/*.jar wdk.jar
+COPY artifact/*.jar wdk.jar
```
yinan-symphony, line 2: after the upload fatjar action, the jar is copied to artifact folder, right?

yinan-symphony (approved):

PR #235, opened 2023-03-07
[CICD] - Added missing java modules to custom jre

symphony-soufiane:
Added missing java modules to the custom jre in order to make execute…-request activity works. Also put default datasource credentials in default application.yaml file.


yinan-symphony (approved):
LGTM

PR #272, opened 2023-05-29
Upgrade Dockerfile to Java 17

### LINKED ISSUE #269: Upgrade Docker Image to Java 17
### Description of Problem:
WDK allows dynamic loading of custom activities and other modules via the `lib` directory. When using the docker image `finos/symphony-wdk`, this defaults to using a Java 11 runtime, which cannot load JARs built with Java 17.

### Potential Solutions:
Upgrade custom JRE used in `Dockerfile` to Java 17

ystxn:
### Description
Closes #269 
Upgrade Dockerfile to Java 17

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,46 +1,11 @@
-FROM openjdk:11-slim-buster AS builder
-COPY artifact/*.jar wdk.jar
+FROM openjdk:17-alpine
+RUN jlink --no-header-files --no-man-pages --compress=2 --strip-java-debug-attributes --output /jre \
+    --add-modules java.base,java.scripting,java.sql,jdk.unsupported,java.se,jdk.crypto.ec
 
-# Create custom JRE
-RUN mkdir custom
-RUN jlink --no-header-files --no-man-pages --compress=2 --strip-debug --add-modules java.base,java.scripting,java.sql,jdk.unsupported,java.se,jdk.crypto.ec --output custom/jre
-
-FROM alpine:3.11
+FROM alpine:3.18
 ENV LANG='en_US.UTF-8' LANGUAGE='en_US:en' LC_ALL='en_US.UTF-8'
+COPY --from=0 /jre /jre
+COPY artifact/*.jar /wdk.jar
+WORKDIR /symphony
 
-RUN apk add --no-cache --virtual .build-deps curl binutils \
-    && GLIBC_VER="2.29-r0" \
-    && ALPINE_GLIBC_REPO="https://github.com/sgerrand/alpine-pkg-glibc/releases/download" \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,46 +1,11 @@
...
-    && tar -xf /tmp/libz.tar.xz -C /tmp/libz \
-    && mv /tmp/libz/usr/lib/libz.so* /usr/glibc-compat/lib \
-    && apk del --purge .build-deps glibc-i18n \
-    && rm -rf /tmp/*.apk /tmp/gcc /tmp/gcc-libs.tar.xz /tmp/libz /tmp/libz.tar.xz /var/cache/apk/*
-
-COPY --from=builder custom/jre /custom/jre
-COPY --from=builder wdk.jar wdk.jar
-
-ENTRYPOINT ["custom/jre/bin/java", "-Dspring.config.additional-location=./symphony/", "-Dloader.path=./symphony/libs", "-jar", "wdk.jar", "--spring.profiles.active=${PROFILE:default}"]
+ENTRYPOINT [ "/jre/bin/java", "-jar", "wdk.jar", "--spring.profiles.active=${PROFILE:default}" ]
```
yinan-symphony, line 11: @ystxn 
additional-location will simplify the application.yml that custom has to config, for example, some internal technical config, we may prefer to leave it in the embedded application.yml. Without additional-location, custom must provide everything.
 
And for loader.path, this will disallow custom to deploy custom activities along side the wdk, is it good ?

Dockerfile
```diff
@@ -1,46 +1,11 @@
-FROM openjdk:11-slim-buster AS builder
-COPY artifact/*.jar wdk.jar
-
-# Create custom JRE
-RUN mkdir custom
-RUN jlink --no-header-files --no-man-pages --compress=2 --strip-debug --add-modules java.base,java.scripting,java.sql,jdk.unsupported,java.se,jdk.crypto.ec --output custom/jre
+FROM openjdk:17-alpine
+RUN jlink --no-header-files --no-man-pages --compress=2 --strip-java-debug-attributes --output /jre \
```
yinan-symphony, line 2: is jdeps necessary before jlink ?

ystxn: only if the deps will change due to new dependencies in the JAR. since wdk is quite stable now, a static list of deps in jlink should work. just need to verify that any major changes in the future doesn't break this.

Dockerfile
ystxn: `loader.path` defaults to ./lib of the working directory, so it will still work

yinan-symphony (approved):

Dockerfile
ystxn: as discussed, `spring.config.additional-location` is optional if the location is already set as the `WORKDIR`. both `application.yaml` and profile-specific e.g. `application-prod.yaml` will work.

